PR #4, opened 2017-05-20
Refactor code

faizan-khan-iit:
I think the current code in the `animint` compiler is a bit hard to debug. Especially because the functions are too long (`saveLayer` spans about 500 lines) and the implementation is mixed up with the logic. As such I sometimes find it difficult to locate the problem as I have to wade through a lot of unrelated code before I get to the buggy part. And often I have a lot of variables that were probably used a 100 lines back cluttering the work-space.

This PR has 2 motives:

- Separate the implementation from the code: The implementation is separated in a file `animintHelpers.R` (for now), while the logic stays in the top level functions. This will avoid us going through trivial code in a debugger that probably is totally unrelated to the problem.

- Reduce the size of functions to easily spot errors and reduce the number of active variables

So the code in the top level functions looks like this:
https://github.com/faizan-khan-iit/animint2/blob/refactor/R/animint.R#L153-L164
with their implementations in other files.

Note: The file `helperFunctions.R` is composed of functions we probably will never need to care about. So I separated those too.


faizan-khan-iit:
@tdhock @cpsievert Please suggest any changes to the approach. Alternatively, we could separate the main functions `animint2dir`, `saveLayer` and `parsePlot` in their own unique files.

R/doc.R
```diff
@@ -0,0 +1,82 @@
+##' Make documentation web site.
+##' @param doc.dir containing example subdirectories.
+##' @return nothing.
+##' @author Toby Dylan Hocking
+##' @export
+##' 
+makeDocs <- function(doc.dir){
```
tdhock: I would suggest deleting this in animint2

tdhock:
I would suggest removing the `meta` environment, if possible. since it is mutable, it also makes the code harder to understand.

faizan-khan-iit:
I too think removing the `meta` environment will help a lot. But it will take some time. I will try it over the weekend and see what I can do..

faizan-khan-iit:
@tdhock Do we need to retain the `gg2animint` function? I think it will be better that we remove it in the new version.

tdhock:
it would be great to remove the gg2animint function

R/animint.R
```diff
@@ -955,7 +950,6 @@ animint2dir <- function(plot.list, out.dir = tempfile(),
   g.list <- list()
   for(p.name in names(ggplot.list)){
     ggplot.info <- ggplot.list[[p.name]]
-    meta$prev.class <- NULL # first geom of any plot should not be next.
     for(layer.i in seq_along(ggplot.info$ggplot$layers)){
```
faizan-khan-iit: Do not need this, I think. It is not updated anywhere in the code

faizan-khan-iit:
@tdhock I was thinking we could merge this PR now, and work on the rest of clean up later. Some part of the code is a bit tricky to deal with, especially selectors that are passed through the `meta` environment. We can sort it out later.

tdhock:
ok

faizan-khan-iit:
Just as a last note, what should we keep the version number? `v2.2017.06.12`?

tdhock:
no need for v2 in front of the version number, but please change the
package name to animint2.

I would recommend to continue using dates for version numbers, e.g.
2017.06.12

PR #5, opened 2017-06-14
ggplot2 dependency

faizan-khan-iit:
The motive of this PR is to solve the `ggplot2` dependency that `animint` had in previous versions.


faizan-khan-iit:
@tdhock Ever since I added the `ggplot2` code, I see this error message while trying to build the package:

```
Error in eval(expr, envir, enclos) : could not find function "grob"
Error : unable to load R code in package 'animint2'
ERROR: lazy loading failed for package 'animint2'
```

I have done the following but the error is recurring:
1. Using collate I tried to shuffle the order of some relevant files but it didn't work. Currently I use the same collate description as [from the ggplot2 package](https://github.com/tidyverse/ggplot2/blob/126b987e7d5d8b846a43d1ed12bc5d927d27af51/DESCRIPTION#L54-L224), and [then load all the animint code](https://github.com/tdhock/animint2/blob/507dae2de6c1c61133eed535b88683b7ac40a6da/DESCRIPTION#L215-L226).
2. I added the Imports and Suggests fields exactly from the ggplot2 code. Still the error persists.

Any ideas??

tdhock:
grob is defined in the grid pacakge, maybe you need to add that to our imports?

faizan-khan-iit:
@tdhock Could you list the downfalls to this approach?

1. We fork `ggplot2`
2. Rename the package to say `ggplot2Animint`
3. Make the changes there and use it for our purposes

I think this way, both of our problems get solved:
1. We don't have to worry about the `ggplot2` version, since users can use the original CRAN version (at least I think its possible)
2. We still have all the freedom, since we won't be sending the `ggplot2` guys any PRs

I was thinking along these lines. Is it possible(to have two versions of `ggplot2` with different names in an environment)? If yes, is it doable?

tdhock:
I don't understand, can you please clarify?

faizan-khan-iit:
The problem earlier was that users could only have one version of `ggplot2`. So our forked version, say v2.1.0, was installed by replacing the CRAN version.

If we change the name to `ggplot2Animint` (say), the user will still have his CRAN version of `ggplot2`. So We could depend on our `ggplot2Animint` fork, so when the user installs `animint2`, `ggplot2Animint` gets installed too. That way, we could solve both the problems. The user can have his updated version of `ggplot2` from CRAN, and we could use the code in `ggplot2Animint` for our purposes. If it is possible, that is.

tdhock:
I guess it is possible, but what is the benefit of having the separate ggplot2Animint package? Why not just put everythign in animint2?

faizan-khan-iit:
Yes, I mean it is technically the same thing, but I think it would be neater to just copy their repo along with the testing framework to make sure we don't break anything critical. Additionally we will also need to add our own tests.

Also, like you mentioned, using the new `ggplot2Animint` package, we could first create static plots before passing them to `animint2`, much like we used to before. I think it will be much more easily done this way.

Using the current approach, there will always be that task of chipping away the useless code; while I think we could just leave that as it is in the other package, without much hassle. Also, we could easily track any additions in the original `ggplot2`.

And finally just some good programming practice to separate the animation code from the static code. Tracking bugs will be easier if I know that the `ggplot2` code is working as it should.

Most of this can be done the current way too. But I would prefer this one. Looks easier and neater.

tdhock:
ok in that case do we change all of geom_ to ageom_ etc? or do we keep geom_?

tdhock:
@cpsievert can you please share your opinion?

faizan-khan-iit:
We will make all the relevant changes in the newer package. For e.g.:
1. We will change the geom names to `geom_a*()`
2. We will not export the functions that are already exported from `ggplot2` and are unchanged
3. Will add the new syntax and tests for allowing the new syntax etc.

tdhock:
for me it seems like making two packages is significantly more complicated
than making one package (two times installation, two times CRAN checks, etc)

to make sure that ggplot2 code we fork is working as it should, we can just
copy their tests to animint2 you know.

BTW, their tests are not nearly as thorough as ours. (they don't have a way
to test the visual properties of the rendered ggplots, which we do by
parsing the generated HTML)

tdhock:
futhermore many of the ggplot2 tests are irrelevant for us. e.g. we won't
use the ggplot_build function anymore because it does not know about
showSelected, so we don't care whether or not it works.

faizan-khan-iit:
> for me it seems like making two packages is significantly more complicated than making one package (two times installation, two times CRAN checks, etc) 

I was kind of thinking the other way around. The `ggplot2` package passes the tests currently and should not have any problems installing, so making changes to allow the new syntax should be easier. Same for `animint2`.

> we won't use the ggplot_build function anymore

Regarding the tests, we could definitely copy them. But from a maintainers perspective, it would be easier with a separate package because:
1. While we may not use some of the code/tests from ggplot2, it will be time consuming to find it and remove it from the package. Let's leave them as they are in the separate package
2. Since most of the relevant code will be in animint, we could separate the geom definitions, themes, and facet computations to the ggplot2Animint package that way

Anyway, it is just bookkeeping and I don't think there are any major advantages/disadvantages. Just my personal viewpoint. The only significant difference will be that if we maintain a separate package, we can use that to create static plots for visual testing first, like you mentioned before. I don't think we will be able to do that if we copy the code(?)

faizan-khan-iit:
@tdhock Should we depend on your fork of `ggplot2`, for consistency?

tdhock:
hey Faizan it seems to be working fine now that I have removed the browser() statements

tdhock:
actually I have to use devtools::load_all() to get it working

tdhock:
```r
> library(animint2)
> g <- ggplot()+geom_point(aes(Sepal.Length, Sepal.Width), data=iris, clickSelects="Species")
> g
> g <- ggplot()+geom_point(aes(Sepal.Length, Sepal.Width), data=iris, clickSelects="Species")
Error in plot_clone(p) : attempt to apply non-function
> devtools::load_all("~/R/animint2")
Loading animint2
> g <- ggplot()+geom_point(aes(Sepal.Length, Sepal.Width), data=iris, clickSelects="Species")
> g
> g <- ggplot()+geom_point(aes(Sepal.Length, Sepal.Width), data=iris, clickSelects="Species")
> g
```

faizan-khan-iit:
That is weird. Could you please try this? I am still getting the same error:
```
mtcars$cyl <- as.factor(mtcars$cyl)

p1 <- ggplot() + geom_point(aes(x=hp,
                                y=disp),
                             showSelected = c("cyl", "am"),
                             clickSelects = c("am"),
                             data = mtcars)

p2 <- ggplot() + geom_point(aes(x=hp,
                                y=disp,
                                showSelected = cyl),
                             data = mtcars)

p3 <- ggplot() + geom_point(aes(x=hp,
                                y=disp),
                             data = mtcars)
```

tdhock:
works for me after load_all:

```r
> library(animint2)
> mtcars$cyl <- as.factor(mtcars$cyl)
> p1 <- ggplot() + geom_point(aes(x=hp,
+                                 y=disp),
+                              showSelected = c("cyl", "am"),
+                              clickSelects = c("am"),
+                              data = mtcars)
Warning: Ignoring unknown parameters: showSelected
> p2 <- ggplot() + geom_point(aes(x=hp,
+                                 y=disp,
+                                 showSelected = cyl),
+                              data = mtcars)
Warning: Ignoring unknown aesthetics: showSelected
> p3 <- ggplot() + geom_point(aes(x=hp,
+                                 y=disp),
+                              data = mtcars)
> p1 <- ggplot() + geom_point(aes(x=hp,
+                                 y=disp),
+                              clickSelects = c("am"),
+                              data = mtcars)
> p3
> p3
Error in plot_clone(plot) : attempt to apply non-function
> devtools::load_all("~/R/animint2")
Loading animint2
> p3
> p1
> 
```

faizan-khan-iit:
Here is what I am doing:
1. Load All
2. Build and Reload
3. Try running the code above

```
> library(animint2)
> mtcars$cyl <- as.factor(mtcars$cyl)
> p1 <- ggplot() + geom_point(aes(x=hp,
+                                 y=disp),
+                              showSelected = c("cyl", "am"),
+                              clickSelects = c("am"),
+                              data = mtcars)
Warning: Ignoring unknown parameters: showSelected
> p2 <- ggplot() + geom_point(aes(x=hp,
+                                 y=disp,
+                                 showSelected = cyl),
+                              data = mtcars)
Warning: Ignoring unknown aesthetics: showSelected
> p3 <- ggplot() + geom_point(aes(x=hp,
+                                 y=disp),
+                              data = mtcars)
> p1
> p2
Error in plot_clone(plot) : attempt to apply non-function
```

tdhock:
right, if you build and reload then you get the error

but if you load_all you dont get the error

here is another hint: you don't get the error if you call animint2:::print.ggplot explicitly. This suggests there is some issue with method dispatch... not sure.

```r
> library(animint2)
> g <- ggplot()
> g
[1] "in print.ggplot at the beginning"
[1] "in print.ggplot before ggplot_build"
[1] "in ggplot_build"
[1] "in plot_clone"
[1] "in clone method"
[1] "in plot_clone"
[1] "in clone method"
> g
Error in plot_clone(plot) : attempt to apply non-function
> print(g)
Error in plot_clone(plot) : attempt to apply non-function
> animint2:::print.ggplot(g)
[1] "in print.ggplot at the beginning"
[1] "in print.ggplot before ggplot_build"
[1] "in ggplot_build"
[1] "in plot_clone"
[1] "in clone method"
[1] "in plot_clone"
[1] "in clone method"
> 
```

tdhock:
another hint:

```r
> g <- ggplot()
> str(g)
List of 9
 $ data       : list()
  ..- attr(*, "class")= chr "waiver"
 $ layers     : list()
 $ scales     :Classes 'ScalesList', 'ggproto', 'gg' <ggproto object: Class ScalesList, gg>
    add: function
    clone: function
    find: function
    get_scales: function
    has_scale: function
    input: function
    n: function
    non_position_scales: function
    scales: NULL
    super:  <ggproto object: Class ScalesList, gg> 
 $ mapping    : list()
 $ theme      : list()
 $ coordinates:Classes 'CoordCartesian', 'Coord', 'ggproto', 'gg' <ggproto object: Class CoordCartesian, Coord, gg>
    aspect: function
    distance: function
    expand: TRUE
    is_linear: function
    labels: function
    limits: list
    modify_scales: function
    range: function
    render_axis_h: function
    render_axis_v: function
    render_bg: function
    render_fg: function
    setup_data: function
    setup_layout: function
    setup_panel_params: function
    setup_params: function
    transform: function
    super:  <ggproto object: Class CoordCartesian, Coord, gg> 
 $ facet      :Classes 'FacetNull', 'Facet', 'ggproto', 'gg' <ggproto object: Class FacetNull, Facet, gg>
    compute_layout: function
    draw_back: function
    draw_front: function
    draw_labels: function
    draw_panels: function
    finish_data: function
    init_scales: function
    map_data: function
    params: list
    setup_data: function
    setup_params: function
    shrink: TRUE
    train_scales: function
    vars: function
    super:  <ggproto object: Class FacetNull, Facet, gg> 
 $ plot_env   :<environment: R_GlobalEnv> 
 $ labels     : list()
 - attr(*, "class")= chr [1:2] "gg" "ggplot"
> g
[1] "in print.ggplot at the beginning"
[1] "in print.ggplot before ggplot_build"
[1] "in ggplot_build"
[1] "in plot_clone"
[1] "in clone method"
[1] "in plot_clone"
[1] "in clone method"
> str(g)
List of 9
 $ data       : list()
  ..- attr(*, "class")= chr "waiver"
 $ layers     : list()
 $ scales     :Classes 'ScalesList', 'ggproto', 'gg' Error in as.vector(x, "list") : 
  cannot coerce type 'closure' to vector of type 'list'
> 
```
Something in the ggplot has been changed by printing it. (I dont think that is normal)

tdhock:
another hint: unloading animint2 and then reloading it solves the problem

```r
> library(animint2)
> g <- ggplot()
> g
[1] "in print.ggplot at the beginning"
[1] "in print.ggplot before ggplot_build"
[1] "in ggplot_build"
[1] "in plot_clone"
[1] "in clone method"
[1] "in plot_clone"
[1] "in clone method"
> g
Error in plot_clone(plot) : attempt to apply non-function
> unloadNamespace("animint2")
> library(animint2)
> g
[1] "in print.ggplot at the beginning"
[1] "in print.ggplot before ggplot_build"
[1] "in ggplot_build"
[1] "in plot_clone"
[1] "in clone method"
[1] "in plot_clone"
[1] "in clone method"
> g
[1] "in print.ggplot at the beginning"
[1] "in print.ggplot before ggplot_build"
[1] "in ggplot_build"
[1] "in plot_clone"
[1] "in clone method"
[1] "in plot_clone"
[1] "in clone method"
> g
[1] "in print.ggplot at the beginning"
[1] "in print.ggplot before ggplot_build"
[1] "in ggplot_build"
[1] "in plot_clone"
[1] "in clone method"
[1] "in plot_clone"
[1] "in clone method"
> 
```

faizan-khan-iit:
So the problem is with the `NAMESPACE`?

tdhock:
not sure, I haven't tried editing NAMESPACE yet

cpsievert:
@faizan-khan-iit your https://github.com/tdhock/animint2/pull/5#issuecomment-309481543 worked fine for me.

As for the separate package discussion, I don't have strong feelings either way. It would be a bit more clear what code is animint-specific if we had a separate package, but I don't see many pros/cons besides that...

faizan-khan-iit:
@cpsievert Did you not get the error I [mentioned afterwards](https://github.com/tdhock/animint2/pull/5#issuecomment-309488166) when trying to display the plots?

tdhock:
@cpsievert to clarify, the problem is that the second time we try to print a ggplot, after attaching animint2, we get `Error: attempt to apply non-function` --- do you have any idea where that is coming from and how to fix it?

cpsievert:
Hmm, I still can't replicate, but when I see that error in plotly it means that the `ggplot_build()` function definition being used internally is different from the (installed) `ggplot2::ggplot_build()` definition.

I suspect what's happening is that, for some reason, `animint2:::print.ggplot` is calling `ggplot2::ggplot_build()` rather than `animint2:::ggplot_build()`

tdhock:
thanks for your expert opinion carson.

I removed ggplot2 from my library, and I got the following error, which confirms that for some reason our code is trying to use the ggplot2 functions.

```
> library(animint2)
> g <- ggplot()
> g
[1] "in print.ggplot at the beginning"
Error in loadNamespace(name) : there is no package called ‘ggplot2’
> 
```

faizan-khan-iit:
[This code](https://github.com/tidyverse/ggplot2/blob/master/R/plot.r#L159-L165) might be the problem here. It loads the Namespace from `ggplot2`, which might be the cause of problems.

tdhock:
definitely, try changing it to animint2

faizan-khan-iit:
I tried changing it, bit it didn't work. I tried reloading the package in a fresh session, but it still produced the same error. So I tried it separately, [in the `ggplot2` fork](https://github.com/tdhock/ggplot2/commit/8738a2e5508bcf28a451da8e0f85cbb457425e3f), but the error still persists.

```
Error in loadNamespace(name) : there is no package called ‘ggplot2’
```

faizan-khan-iit:
@tdhock Finally some progress. I am currently working with the [`validate-params` branch in the my fork](https://github.com/faizan-khan-iit/ggplot2/tree/validate-params), and apart from some warnings (which I guess are due to some internal R changes), the new syntax seems to work. I changed the name of the package in my branch to `ggplot2Animint`, and this is what I currently depend on in the `animint2` code.

I have only added the new syntax support, and these two examples seem to be identical to the older syntax
```
## Example 1
mtcars$cyl <- as.factor(mtcars$cyl)

# Older syntax
p1 <- ggplot() + geom_point(aes(x=hp,
                                y=disp,
                                showSelected1 = cyl,
                                showSelected2 = am),
                             data = mtcars)
# New syntax
p2 <- ggplot() + geom_point(aes(x=hp,
                                y=disp),
                            showSelected = c("cyl", "am"), 
                            data = mtcars)

animint2::animint2dir(list(p1=p1, p2=p2), "~/test_animint2_initial2")

## Example 2
# Old syntax
p4 <- ggplot() + geom_point(aes(x=conc,
                                y=uptake,
                                showSelected1 = Treatment,
                                showSelected2 = Type),
                            data = CO2)
# New syntax
p5 <- ggplot() + geom_point(aes(x=conc,
                                y=uptake),
                            showSelected = c("Treatment", "Type"),
                            data = CO2)

animint2::animint2dir(list(p4=p4, p5=p5), "~/test_animint2_initial4")
```

I will need to make some more changes to the `animint2` code to add proper support. Does this look okay?

tdhock:
these examples work on my computer, Ubuntu 14.04 with R-3.4.1

tdhock:
hey I converted clickSelects/showSelected in the compiler tests from aes to params.

I also have turned on travis and tests.

So far I have not yet modified key/tooltip aes -- do you think we should change them to params as well or keep as is?

faizan-khan-iit:
> key/tooltip aes -- do you think we should change them to params as well or keep as is

I guess we could leave them be for now. But after we get the newer syntax working, we could consider these then.

On a side note, for the `showSelected` value added automatically added when we use `colour` aesthetic, do you think we should support that?

tdhock:
ok to leave key/tooltip for now.

in the old animint we added showSelectedColour aes when colour aes was present, and it generated an interactive legend.

in the new animint if there is aes(colour=Species) I would suggest that by default the compiler adds Species to the character vector of showSelected values.

tdhock:
hey @faizan-khan-iit  in our skype yesterday I was saying that we should not add clickSelects/showSelected to the aes anymore, but after thinking about it, maybe I was wrong. I guess we will need to have those variables in the tsv files eventually, and maybe the easiest way to do that is by adding them as aes (programmatically, based on the new params syntax). It still would probably be a good idea to think about removing the selector.aes function (and replacing it with something simpler that just looks at the clickSelects/showSelected params). What do you think?

faizan-khan-iit:
@tdhock So regarding the current changes that I have made, is there anything you would like to change (apart from removing `selector.aes`, I agree that should be removed)?

As for the `selector.aes`, I was looking at the tests files for .variable and .value, [I guess we don't need these anymore](https://github.com/tdhock/animint2/blob/60aeef024d0e60ea64febc5e7e3b7ac4ec1a8485/tests/testthat/test-renderer1-variable-value.R#L3-L31). But what I wanted to ask is the syntax for [this example](https://github.com/tdhock/animint2/blob/60aeef024d0e60ea64febc5e7e3b7ac4ec1a8485/tests/testthat/test-renderer1-variable-value.R#L99-L100). Could you specify that? Either in the file, I won't make any changes to the file to avoid conflicts, or just comment the syntax here.

tdhock:
nothing else to change for now I think -- your current changes are fine.

For the selector.aes tests you can delete them yes.

For that example we have `showSelected.variable=paste0(problem.name, "peaks"), showSelected.value=peaks, showSelected2=bases.per.problem` and the new syntax should be 
```r
showSelected.vec <- c("bases.per.problem", "peaks")
names(showSelected.vec) <- c("", paste0(problem.name, "peaks"))
geom_segment(showSelected=showSelected.vec)
```
Does that make sense?

faizan-khan-iit:
@tdhock I had to make some changes to the syntax you proposed. That syntax gives an error while declaring the `showSelected.vec` because the `names` attribute has more elements than the vector. 

```
paste0(problem.name, "peaks")
```

Evaluates as an expression and we have more name attributes than the vector length

faizan-khan-iit:
So I edited the syntax to this:
```
showSelected.vec <- c("bases.per.problem",
                      as.var(c(paste0(peak.problems$problem.name,
                                      "peaks"))),
                      as.val("peaks"))
```

where the function `as.var` and `as.val` are as follows:
```
as.var <- function(cs_or_ss_variable){
  return(c(variable = lazyeval::expr_text(cs_or_ss_variable)))
}

as.val <- function(cs_or_ss_value){
  return(c(value = cs_or_ss_value))
}
```

faizan-khan-iit:
So basically, the `showSelected.variable` will be passed using `as.var` (which we will export) and the same for `showSelected.value`. `clickSelects` will mirror the same approach

faizan-khan-iit:
Does this look fine? Basically I did not want the `showSelected.variable` to be evaluated before passing to the animint code so I converted the expression to text using `expr_text` and wrapping it in `as.var` for better syntax

tdhock:
Hey so can you explain again why the syntax I proposed doesn't work? It
seems to me like it is a lot simpler to just used a named character vector,
instead of introducing two new functions (as.var, as.val).

faizan-khan-iit:
In the syntax, if I got it right, it would be like this:
```
showSelected.vec <- c("bases.per.problem", "peaks")
names(showSelected.vec) <- c("", paste0(peak.problems$problem.name, "peaks"))
```

In general, for some data frame `data_frame` we would have something like:
```
showSelected.vec <- c("A", "some_col_from_data_frame")
names(showSelected.vec) <- c("", paste0(data_frame$some_other_col, "foo"))
```
Here the length of `showSelected.vec ` is 2, but the expression `paste0(data_frame$some_other_col, "foo")` will be evaluated according to `data_frame`, so the length of the `names` we are trying to assign will change, (won't it?) Say `data_frame` has 20 rows, then the names attribute will be of length 21. So this will give an error. 

Did I get this right?

tdhock:
not that is not right, but it is my fault for not being clear. here is my proposition. for example we have 
```r
> head(peak.problems)
                   problemStart problemEnd problem.i bases.per.problem
size.100.problem.1          100        200         1               100
size.100.problem.2          200        300         2               100
size.50.problem.1           100        150         1                50
size.50.problem.2           150        200         2                50
size.50.problem.3           200        250         3                50
size.50.problem.4           250        300         4                50
                         problem.name peakStart peakEnd peaks
size.100.problem.1 size.100.problem.1       110     190     1
size.100.problem.2 size.100.problem.2       210     290     1
size.50.problem.1   size.50.problem.1       110     140     1
size.50.problem.2   size.50.problem.2       160     190     1
size.50.problem.3   size.50.problem.3       210     240     1
size.50.problem.4   size.50.problem.4       260     290     1
> 
```
currently we have 
```r
geom_segment(aes(
                          showSelected.variable=paste0(problem.name, "peaks"),
                          showSelected.value=peaks,
                          clickSelects=problem.name,
                          showSelected2=bases.per.problem))
```
- The clickSelects refers to a selector named `problem.name` with values given in the problem.name column.
- The showSelected.variable here refers to selectors with variable names given in the problem.name column, e.g. `size.100.problem.1peaks` and `size.50.problem.1peaks`, etc. 
- The showSelected.value here refers to selector values given in the peaks column e.g. 1 above.

For this example I would propose the new syntax
```r
geom_segment(
 showSelected=c(problem.name="peaks", "bases.per.problem"),
 clickSelects="problem.name")
```

tdhock:
Note that the showSelected.variable names are different in this new syntax, e.g. size.100.problem.1 rather than size.100.problem.1peaks -- this is fine in my opinion. To exactly replicate the previous example we would need to create a new column in the peak.problems data table, 
```r
> peak.problems$selector.name <- with(peak.problems, paste0(problem.name, "peaks"))
> rownames(peak.problems) <- NULL
> head(peak.problems)
  problemStart problemEnd problem.i bases.per.problem       problem.name
1          100        200         1               100 size.100.problem.1
2          200        300         2               100 size.100.problem.2
3          100        150         1                50  size.50.problem.1
4          150        200         2                50  size.50.problem.2
5          200        250         3                50  size.50.problem.3
6          250        300         4                50  size.50.problem.4
  peakStart peakEnd peaks           selector.name
1       110     190     1 size.100.problem.1peaks
2       210     290     1 size.100.problem.2peaks
3       110     140     1  size.50.problem.1peaks
4       160     190     1  size.50.problem.2peaks
5       210     240     1  size.50.problem.3peaks
6       260     290     1  size.50.problem.4peaks
> 
```
Then do 
```r
geom_segment(
 showSelected=c(selector.name="peaks", "bases.per.problem"),
 clickSelects="problem.name")
```

faizan-khan-iit:
Sorry I misunderstood that. I thought we had to keep the `showSelected.variable` names same as before, so I thought the `paste` function was necessary here. Since it was getting evaluated before passing the params, I was getting an error with the lengths. Thanks for the clarification. This will be much neater.

faizan-khan-iit:
@tdhock Could you convert [this part of the variable value test](https://github.com/tdhock/animint2/pull/5/commits/db237520a5fe527f5892af73e5dfdd863e542b57#diff-925d87603006c68c6c866aef22fcb5f6R184) to the newer syntax? I have converted all the rest.

tests/testthat/test-renderer4-prostateLasso.R
```diff
@@ -30,13 +30,11 @@ viz.no.time <- list(
...
-      showSelected.variable="arclength",
-      showSelected.value=arclength.show,
       key=ifelse(
         arclength.click==arclength.show, 1,
         paste(arclength.click, arclength.show))),
+      clickSelects=c(arclength="arclength.click"),
+      showSelected=c(arclength="arclength.show"),
       alpha=0.5,
       data=addY(tallrect.all, "error")),
   res=ggplot()+
```
faizan-khan-iit: @tdhock Here when I converted to the newer syntax, I noticed the data `addY(tallrect.all, "error")`, has no `arclength` column and hence we [encounter this error in the tests](https://travis-ci.org/tdhock/animint2/jobs/264712918#L2955-L2958). I guess just adding the `arclength` column separately should solve this, but I think I should add a separate test to ensure this throws an error in our code. Any comments on this?

tdhock: interesting example. in the old syntax we had 
```r
clickSelects.variable="arclength",
clickSelects.value=arclength.click,
showSelected.variable="arclength",
showSelected.value=arclength.show,
```
and you're right with your new syntax, our compiler will look for a column named arclength. Right now you're right the most direct translation of this example would be to just create a new "selector.name" column in the data table (every row has value "arclength").

Ideally  not right now, but one day as an optimization) we could add a special case for this, I added a TODO https://github.com/tdhock/animint2/commit/5437ab74eec197dc07c2861a2bccdd1da90c6c7e

tdhock:
For the `viz.for` test in variable-value.R, I think your translation is fine (it is supposed to use the non-variable-value selectors).

faizan-khan-iit:
@tdhock Most of the test cases are passing now, including line path and stat summary. But I ran into a problem. In the animint code, we modify the original plot_list and use that for computation, so adding showSelected from params to aesthetics creates a problem during checks. I will explain:

```
data(iris)
viz <- list( length=ggplot()+
                  geom_point(aes(Sepal.Length, Petal.Length),
                  showSelected=c("Species"),
                  data=iris))

## Viz is created, mapping is fine
viz$length$layers[[1]]$mapping
* x -> Sepal.Length
* y -> Petal.Length

animint2dir(viz, "~/test_iris", open.browser = F)      ## First run, works fine

## Animint has added the aesthetics to the viz itself, so the original viz gets edited
viz$length$layers[[1]]$mapping
* x             -> Sepal.Length
* y             -> Petal.Length
* showSelected1 -> Species      --->>>   ### Added by compiler ###

## On second run, we already have showSelected aesthetics,
## but our compiler doesn't know these were added later on
## So it throws an error assuming the user specified them 
animint2dir(viz, "~/test_iris", open.browser = F)

Error: Use of clickSelects and showSelected as aesthetics has been deprecated.
Please use as parameters. Problem: 
Plot:  length 
```

This is why I added this code to [first save the original aesthetics mapping](https://github.com/tdhock/animint2/pull/5/commits/5234849c2b445f356f374a927a2bb324b3139944#diff-fddb8085c2de24f7941a89d932954d47R919), and [replace it after everything is finished](https://github.com/tdhock/animint2/pull/5/commits/5234849c2b445f356f374a927a2bb324b3139944#diff-fddb8085c2de24f7941a89d932954d47R1295).

But this is very bad fix and won't work for some cases too, [for e.g. try running multiple update axes tests](https://github.com/tdhock/animint2/blob/5234849c2b445f356f374a927a2bb324b3139944/tests/testthat/test-renderer4-update-axes-multiple-ss.R#L38-L72).

The straight forward way would be to copy the plot in parsePlot instead of editing the original plot. Could you help a bit here?

faizan-khan-iit:
This is especially troublesome when we have definitions in tests where the base plot code is the same and we are only testing animations or updates and stuff.

E.g.
```
viz.no.duration <- list(...##define plot here##...)
animint2dir(viz.no.duration)

viz.duration <- viz.no.duration
## Add animation details
animint2dir(viz.duration)      ---->>>> ## throws error
```

tdhock:
thats interesting, I did not know that editing the aes object in a function would result in changes after that function is done. But this is reproducible on my machine as well:
```r
> library(ggplot2Animint)
> g <- ggplot()+geom_point(aes(Petal.Length, Sepal.Length), data=iris)
> f <- function()g$layers[[1]]$mapping$showSelected <- as.symbol("val")
> g$layers[[1]]$mapping
* x -> Petal.Length
* y -> Sepal.Length
> f()
> g$layers[[1]]$mapping
* x            -> Petal.Length
* y            -> Sepal.Length
* showSelected -> val
> 
```
even stranger, if you create another ggplot, then edit the mapping in the original, the mapping in the derived version changes as well:
```r
> g <- ggplot()+geom_point(aes(Petal.Length, Sepal.Length), data=iris)
> g.copy <- g+geom_blank()
> f()
> g.copy$layers[[1]]$mapping
* x            -> Petal.Length
* y            -> Sepal.Length
* showSelected -> val
> 
```
same thing if you just copy the layers:
```r
> g <- ggplot()+geom_point(aes(Petal.Length, Sepal.Length), data=iris)
> layers.copy <- as.list(g$layers)
> layers.copy[[1]]$mapping
* x -> Petal.Length
* y -> Sepal.Length
> f()
> layers.copy[[1]]$mapping
* x            -> Petal.Length
* y            -> Sepal.Length
* showSelected -> val
> 
```
one solution could be to copy the mapping before editing it, e.g. 
```r
> g <- ggplot()+geom_point(aes(Petal.Length, Sepal.Length), data=iris)
> g$layers[[1]]$mapping.orig <- g$layers[[1]]$mapping
> f()
> g$layers[[1]]$mapping.orig
* x -> Petal.Length
* y -> Sepal.Length
> g$layers[[1]]$mapping
* x            -> Petal.Length
* y            -> Sepal.Length
* showSelected -> val
> 
```
then copy it back at the end, e.g.
```r
> g$layers[[1]]$mapping <- g$layers[[1]]$mapping.orig
> g$layers[[1]]$mapping
* x -> Petal.Length
* y -> Sepal.Length
> 
```
any other ideas?

R/animint.R
```diff
@@ -962,26 +938,47 @@ animint2dir <- function(plot.list, out.dir = tempfile(),
...
+  ## Call ggplot_build in parsPlot for all ggplots
+  ggplot.list <- list()
+  AllPlotsInfo <- list()
+  for(list.name in names(plot.list)){
+    p <- plot.list[[list.name]]
+    if(is.ggplot(p)){
+      ## If plot is correct, save to meta for further processing
+      parsed_info <- parsePlot(meta, p, list.name) # calls ggplot_build.
+      AllPlotsInfo[[list.name]] <- parsed_info[[1]]
+      ggplot.list[[list.name]] <- parsed_info[2:3]
```
tdhock: it would be easier to read/understand if you use `$name` rather than `[[1]]` or `[2:3]`

tdhock:
I fixed the knitr test by changing knit_print.animint2 to knit_print.animint (the S3 method was not being called, so the default print method -- static ggplots -- was being used in the Rmd rendering process)

faizan-khan-iit:
@tdhock We are passing all the tests now except the compiler-gist test. So I guess we could start thinking about closing this PR. Is there anything else that can be done here?

tdhock:
Hey Faizan this is great news, thanks for all your hard work!
I would say let's fix the gist test (finally) before merging this PR. 

I just committed the secure string to the .travis.yml file, which I hope will fix the gist issue. otherwise maybe we need to generate a new gist api key and store it in a new travis secure https://docs.travis-ci.com/user/encryption-keys/

faizan-khan-iit:
@tdhock Thanks!

I have added the examples to the package and converted them to the new syntax but I haven't tested them yet. Maybe we could do that in a separate PR?

Also looks like we will have to generate a new gist api key..

tdhock:
hey so I generated a new gist access token under github.com/settings, then I did `travis encrypt GITHUB_PAT="ACCES_TOKEN_VALUE"` which got me a new secure string that I put in .travis.yml, hopefully this solves the gist issue.

tdhock:
ok, all tests are now passing, and I have updated the version in NEWS/DESCRIPTION/README. I give my official +1 for this PR, go ahead and merge when you are ready.

PR #8, opened 2017-08-28
Use data.table package in compiler

faizan-khan-iit:
Optimize the code using the `data.table` package. Currently we use base R functions, particularly for writing files

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,6 +20,7 @@ Imports:
     stats,
     tibble,
     lazyeval,
+    data.table,
 Suggests:
     knitr (>= 1.5.33),
     servr,
--- a/R/animint.R
+++ b/R/animint.R
@@ -627,7 +627,7 @@ saveLayer <- function(l, d, meta, layer_name, ggplot, built, AnimationInfo){
       ## chunks if there are any that are too small.
       tmp <- tempfile()
       some.lines <- rbind(head(g.data), tail(g.data))
-      write.table(some.lines, tmp,
+      data.table::fwrite(some.lines, tmp,
                   col.names=FALSE,
                   quote=FALSE, row.names=FALSE, sep="\t")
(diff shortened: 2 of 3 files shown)


faizan-khan-iit:
@tdhock I was looking at the code that deals with transformation of data in `split.x` and `getCommonChunk`. One way to use the `data.table` package would be to convert the `data.frame` used currently using `as.data.table(data_frame)`. Then most of the methods like `split` and and `na.omit` will be used from the `data.table` package.

After that we could consider the for loops..

tdhock:
I would say before attempting any conversions, we should try to run a
profiler to make sure that a substantial amount of time is being spent in
getCommonChunk. If so, then re-writing getCommonChunk using data.table()[,
by=chunk.vars] instead of split would be a good idea. If not, it would be
better to focus our efforts on other sub-routines that take more time.

faizan-khan-iit:
I did that, and most of the time is actually spent in `getCommonChunk`. I ran the WorldBank example 10 times over and here is the result in brief (is there a way to share the full profvis result file??):

| animint2dir -> 27.6 s
|
| | | storeLayer -> 1.5 s
|
| | | saveLayer -> 19.5 s
| | | | | | getCommonChunk  -> 11.6 s
| 
| | |parsePlot -> 5.5 s

So I guess we could assume that getCommonChunk needs some work.

tdhock:
ok great, in that case let's focus on getCommonChunk.

not sure about how to share the profvis results, but maybe a screenshot?

faizan-khan-iit:
1. `saveLayer` takes most time in `animint2dir`
no_table1

------------------------------------------

2. `getCommonChunk` and an `apply` function take time in `saveLayer`
no_table2

 --------------------------------------------- 

3. `getCommonChunk`
no_table3


------------------------------------------

4. `sapply`
no_table4

PR #15, opened 2018-06-02
get_alpha_stroke status and questions

adrianblanco:
I am working during the weekend so probably I will update this PR later
Status and questions so far


inst/htmljs/animint.js
```diff
@@ -27,13 +27,13 @@ var animint = function (to_select, json_file) {
             d[v_name] = parseInt(d[v_name]);
           } else if (r_type == "numeric") {
             d[v_name] = parseFloat(d[v_name]);
-          } else if (r_type == "factor" || r_type == "rgb" 
-		     || r_type == "linetype" || r_type == "label" 
+          } else if (r_type == "factor" || r_type == "rgb"
+		     || r_type == "linetype" || r_type == "label"
```
tdhock: please do not commit anything like this which shows up in the diff, but actually does not affect anything other than the formatting. (it makes it more difficult to see where the real changes are in the diff)

tdhock:
In this PR you have modified the animint.js code.

Instead of modifying the animint.js code at first, I would really encourage you to use test-driven development: start by adding a new test, and then modify the animint.js code after that.

PR #16, opened 2018-06-06
get_alpha_stroke test

adrianblanco:
- [x] New test added. It pass as the parameters are fixed with validate_params = FALSE.
- [x] Working on a complimentary test which contains a cartogram. The idea is to test the alpha stroke in the borders of the map.
- [ ] Next step: modify the animint.js code after that. Any advice, recommendations about how should I proceed?


adrianblanco:
I updated the test with a cartogram. I did not create a new PR in order not to open several at the same time. Here it is the link to the test: 
https://github.com/tdhock/animint2/blob/GSOC18/tests/testthat/test-renderer1-get_alpha_stroke.R

tdhock:
did you change anything other than adding the test-renderer1-get_alpha_stroke.R file? I can't easily tell because there are 218 files changed!

I assume you used `git commit -a` ? It would be preferable to not use `-a` and instead commit only the files you have changed (makes it much easier to review). In any case please change the file permissions back, so the "Files changed" tab here only shows the actual changes you have made.

adrianblanco:
Thanks for the feedback.

I used Desktop for that commit and I messed it up. I only made changes in test-renderer1-get_alpha_stroke.R 
I tried to change file permissions following these steps but it is not working for me:
https://stackoverflow.com/questions/2517339/how-to-recover-the-file-permissions-to-what-git-thinks-the-file-should-be
Sorry, I am not familiar with it. From now, I will always use the command line.

- [ ] expectations added. Review pending, not sure if i did it in the right way.
https://github.com/tdhock/animint2/pull/16/commits/d53e9df78b9703129a59873d3fecf4ea280280f2
- [ ] working on another with a dataset already included in animint2

tdhock:
also I noticed that due to some changes in ggplot2Animint, the travis tests were not working correctly (lots of false positive test failures here).

however I fixed that on master via #17  -- hopefully those tests should pass for Travis PR builds when you push a new commit

tdhock:
finally the branch name "GSOC18" is OK for this PR, but in the future please use branch names that reflect the feature that you are working on, as the PR title, e.g. stroke-opacity-params

adrianblanco:
Working on a HTML parsing function to extract the stroke-opacity style and trying to understand the process.

adrianblanco:
Thanks for the advice and for fixing Travis tests. 

Ok, from now I will be more careful choosing names and naming branches. Understood!

adrianblanco:
Working on that. I am taking `var get_alpha` from animint.js as an example

PR #18, opened 2018-06-09
Merge ggplot2Animint into animint2

vivekktiwari:
This PR is to discuss the possibility of merging `ggplot2Animint` into `animint2` to get rid of the dependency. 

For Now: Merging in `dev` branch to avoid conflict 

- [x] Rename files of original animint2 with `a_*` for quick identification. 

- [x] Add ggplot2Animint files

- [x] Add collate to DESCRIPTION of animint2

- [x] Make syntax changes to ggplot2Animint files

- [x] Rename files again of original animint2 from `a_*` to `z_*` to form correct collate

- [x] adding data and data-raw from ggplot2Animint for examples mentioned in code to run
- [ ] Make syntax changes to `z_*` files to remove dependency

Update Tasks: 

- [ ] write tests to make ggplot2 and animint2 work when loaded. 

- [ ] run less tests during CRAN check


vivekktiwari:
@tdhock please provide suggestions on what should be avoided and what to keep in mind when going further in merging both.

tdhock:
The number one priority should be that all tests pass. 
- all existing tests
- add a test to make sure ggplot2 and animint2 can be attached at the same time, probably in a test-compiler-* file?
- it would be great if you could add another entry to the build matrix that runs CRAN checks. During CRAN checks of animint2, we can't run all of our tests, so you need to edit testthat.R so that we have some code that does "if on CRAN, run only a few checks, otherwise run as usual"

vivekktiwari:
@tdhock thank you! Once all the examples pass the `check` function, only then I can make sure the tests will work, right? 

So, I am facing a very typical problem, there are a lot of functions inside the code of ggplot2Animint and animint2 that are same. 

For example: 
1. `train_layout` in `R/panel.r` and in `R/z_facet.r`
2. `build_strip` in `R/facet-labels` and in `R/z_facet.r`

So when running examples, functions are not referencing where they should as both are loaded in the environment, and whoever is loaded at last gets the priority and by build, animint2 code is loaded at last, so gets the override. 

I am sure there are more, what do you recommend for this?

vivekktiwari:
I am getting a single error in Travis build in `compiler`: 
`── 1. Error: animint2gist() returns an object of class 'gist' (@test-compiler-gist.R#4) 
In non-interactive environments, please set GITHUB_PAT env to a GitHub access token (https://help.github.com/articles/creating-an-access-token-for-command-line-use)`

please suggest changes to correct this? Don't know much about access tokens. 

> it would be great if you could add another entry to the build matrix that runs CRAN checks. During CRAN checks of animint2, we can't run all of our tests, so you need to edit testthat.R so that we have some code that does "if on CRAN, run only a few checks, otherwise run as usual

Can I use `skip_on_cran` function for this? please let me know which test units are to be excluded, and will make the necessary changes.

tdhock:
if the functions are really the same then just use one or the other, it
doesn't matter, right?

vivekktiwari:
@tdhock they have different functions/objective, just the names are the same. 

> ── 1. Error: animint2gist() returns an object of class 'gist' (@test-compiler-gist.R#4) In non-interactive environments, please set GITHUB_PAT env to a GitHub access token (https://help.github.com/articles/creating-an-access-token-for-command-line-use)

If you can kindly help me with GITHUB_PAT, my access token keys don't seem to be working on this repo or I might not be doing it correctly.

tdhock:
hi Vivek to avoid the gist error, please send your PR from a branch of
tdhock/animint2 (instead of your fork, vivek/animint2)

I added you as a collaborator on tdhock/animint2 so you should now be able
to do that.

please do not push anything to master

please create a new branch and then a PR from that branch

vivekktiwari:
Thank you so much @tdhock for adding me here. I have created `merge` branch and merging this PR with that. 
I will open PR from this branch and commit changes here.

PR #29, opened 2019-08-15
fix issue with gg+guides(color="none") 

tdhock:


tdhock:
it has something to do with mapping being mutable.

PR #33, opened 2019-11-09
Update README.org

dA505819:
Corrected only spacing.


tdhock:
thanks for the "correction" but it is important to keep "showSelected" because that is the syntax (no spaces) of the geom argument/parameter used for interaction in animint2.

dA505819:
:)

PR #35, opened 2020-03-12
HitOrMissMonteCarlo Test Added with Minor change in timeout setting

lazycipher:
I've added HitOrMissMonteCarlo Test for Hard Test of `Animated interactive ggplots`
I had to change the setImplicitTimeout method as it wasn't working and seems to be defunct.

### DIFF
--- a/tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
+++ b/tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
@@ -90,51 +90,6 @@ viz <- HitOrMissMonteCarlo()
 
 info <- animint2HTML(viz)
 
-# Test of needed data columns in TSV files
-chunk1.tsv <- file.path("animint-htmltest", "geom1_path_vizone_chunk1.tsv")
-chunk1 <- read.table(chunk1.tsv, sep="\t", header=TRUE,
-                     comment.char="", quote="")
-
-chunk2.tsv <- file.path("animint-htmltest", "geom2_point_vizone_chunk1.tsv")
-chunk2 <- read.table(chunk2.tsv, sep="\t", header=TRUE,
-                     comment.char="", quote="")
-
-chunk3.tsv <- file.path("animint-htmltest", "geom5_path_viztwo_chunk1.tsv")
-chunk3 <- read.table(chunk3.tsv, sep="\t", header=TRUE,
-                     comment.char="", quote="")
-
-chunk4.tsv <- file.path("animint-htmltest", "geom6_text_viztwo_chunk1.tsv")
-chunk4 <- read.table(chunk4.tsv, sep="\t", header=TRUE,
-                     comment.char="", quote="")
-
-chunk5.tsv <- file.path("animint-htmltest", "geom4_hline_viztwo_chunk1.tsv")
-chunk5 <- read.table(chunk5.tsv, sep="\t", header=TRUE,
-                     comment.char="", quote="")
-
-chunk6.tsv <- file.path("animint-htmltest", "geom3_tallrect_viztwo_chunk1.tsv")
-chunk6 <- read.table(chunk6.tsv, sep="\t", header=TRUE,
-                     comment.char="", quote="")
-
-test_that("Integration Curve data contains expected columns", {
-  path_expected_cols <- c("x", "y", "group")
-  point_expected_cols <- c("shape", "colour", "x", "y", "key", "showSelectedlegendshapecolour", "showSelected1", "fill")
-  
-  expect_identical(sort(names(chunk1)), sort(path_expected_cols))
-  expect_identical(sort(names(chunk2)), sort(point_expected_cols))
-})
-
-test_that("Pi estimation graph data contains expected columns", {
-  path_expected_cols <- c("x", "y", "showSelected1", "key", "group")
-  text_expected_cols <- c("x", "y", "key", "label", "showSelected1", "group")
-  tallrect_expected_cols <- c("xmin", "xmax", "clickSelects", "group")
-  hline_expected_cols <- c("yintercept")
-  
-  expect_identical(sort(names(chunk3)), sort(path_expected_cols))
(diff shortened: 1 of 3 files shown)


lazycipher:
@tdhock @faizan-khan-iit I've added a test for Hit or Miss Monte Carlo Integration and Pi value estimation.
Please let me know if any changes are required!

tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
```diff
@@ -0,0 +1,171 @@
...
+  expect_identical(sort(names(chunk1)), sort(path_expected_cols))
+  expect_identical(sort(names(chunk2)), sort(point_expected_cols))
+})
+
+test_that("Pi estimation graph data contains expected columns", {
+  path_expected_cols <- c("x", "y", "showSelected1", "key", "group")
+  text_expected_cols <- c("x", "y", "key", "label", "showSelected1", "group")
+  tallrect_expected_cols <- c("xmin", "xmax", "clickSelects", "group")
+  hline_expected_cols <- c("yintercept")
+
```
tdhock: these expected columns tests are not appropriate, because the names/ordering of columns in csv files is a compiler implementation detail, and the test-renderer* files should test visual/renderer functionality. please remove.

tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
```diff
@@ -0,0 +1,171 @@
...
+  
+  expect_equal(length(xlabel), 2)
+  expect_equal(length(ylabel), 2)
+})
+
+test_that("Different Points Are Rendered on Integration Curve", {
+  x1_nodes <- getNodeSet(info$html, "//g[@class='geom2_point_vizone']/g[@class='PANEL1']/circle/@cx")
+  y1_nodes <- getNodeSet(info$html, "//g[@class='geom2_point_vizone']/g[@class='PANEL1']/circle/@cy")
+  
+  Sys.sleep(1.739)
```
tdhock: why this number ?

tdhock:
great start but would be even better if you could edit the .travis.yml config to fix the build system error. I reported the issue here https://github.com/ropensci/RSelenium/issues/230 but no response. maybe you could try upgrading the version of RSelenium used in DESCRIPTION?

tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
lazycipher: Okay got it.
Let me remove this in the next commit.

tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
lazycipher: I took the reference from other successful tests. `Sys.sleep()` is used to pause the program for the specified time so that next reading can be taken. I used the same time as in other tests.

lazycipher:
Okay, sure thing!
I'll upgrade the RSelenium version. There have been some breaking changes from the currently used version and the latest version. I'll take a look and make it work.

tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
```diff
@@ -0,0 +1,171 @@
...
+test_that("Different points are rendered on Pi estimation plot", {
+  coordPanel1 <- getNodeSet(info$html, "//g[@class='geom5_path_viztwo']/g[@class='PANEL1']/path/@d")
+
+  Sys.sleep(1.739)
+
+  info$newHtml <- getHTML()
+
+  coordPanel2 <- getNodeSet(info$newHtml, "//g[@class='geom5_path_viztwo']/g[@class='PANEL1']/path/@d")
+
+  expect_false(identical(coordPanel1, coordPanel2))
```
fzkhan: Does this test assume that `getNodeSet` will always return the rendered points in the same order? I would prefer not to check it this way because even if the plot is not rendering any new points if getNodeSet returns them in a different order, this test might pass. Though if you are sure about the ordering, no changes needed. To me, the length of the retrieved nodes is a better indicator.

lazycipher: I guess checking the length of retrieved nodes is a better indicator. I'll make changes and push as soon as possible.

tdhock:
typically i the order does not matter then you can sort the

tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
lazycipher: Hii @tdhock, @faizan-khan-iit, 
I observed that the points are being rendered in order. 
I checked and found this about the order.
Screenshot from 2020-04-14 13-29-29
Screenshot from 2020-04-14 13-30-17

So, the identical check is good to go. But if you say I can move to length check which will also work!

tests/testthat/test-renderer3-HitOrMissMonteCarlo.R
lazycipher: I've removed these checks.

lazycipher:
> great start but would be even better if you could edit the .travis.yml config to fix the build system error. I reported the issue here [ropensci/RSelenium#230](https://github.com/ropensci/RSelenium/issues/230) but no response. maybe you could try upgrading the version of RSelenium used in DESCRIPTION?

@tdhock, Looks like the error didn't occur this time which occurred last time. But different error occurred this time which is in renderer5 and compiler. Have a look please? https://travis-ci.org/github/tdhock/animint2/builds/674748257
```
The command "Rscript -e 'deps <- remotes::dev_package_deps(dependencies = NA);remotes::install_deps(dependencies = TRUE);if (!all(deps$package %in% installed.packages())) { message("missing: ", paste(setdiff(deps$package, installed.packages()), collapse=", ")); q(status = 1, save = "no")}'" failed and exited with 1 during .
```

PR #45, opened 2020-08-14
Move geom-* specific calculations to geom class definitions.

lazycipher:
This PR moved geom-* specific calculations to geom class definitions.


lazycipher:
@tdhock, All tests are passing now! Should we move ahead with this?

NAMESPACE
```diff
@@ -126,6 +126,7 @@ export(GeomAbline)
 export(GeomAnnotationMap)
 export(GeomArea)
 export(GeomBar)
+export(GeomBind2d)
```
tdhock, line 129: typo?

NAMESPACE
```diff
@@ -137,6 +138,7 @@ export(GeomDensity2d)
 export(GeomDotplot)
 export(GeomErrorbar)
 export(GeomErrorbarh)
+export(GeomFreqploy)
```
tdhock, line 141: typo?

R/geom-bin2d.r
```diff
@@ -31,7 +31,7 @@ geom_bin2d <- function(mapping = NULL, data = NULL,
     data = data,
     mapping = mapping,
     stat = stat,
-    geom = GeomTile,
+    geom = GeomBind2d,
```
tdhock, line 34: why do you change this?

R/geom-bin2d.r
```diff
@@ -41,3 +41,33 @@ geom_bin2d <- function(mapping = NULL, data = NULL,
     )
   )
 }
+
+#' @rdname animint2-gganimintproto
+#' @format NULL
+#' @usage NULL
+#' @export
+#' @include geom-rect.r
+GeomBind2d <- gganimintproto("GeomBind2d", GeomRect,
```
tdhock, line 50: why do you need to add this?

R/geom-freqpoly.r
```diff
@@ -16,10 +16,18 @@ geom_freqpoly <- function(mapping = NULL, data = NULL,
     data = data,
     mapping = mapping,
     stat = stat,
-    geom = GeomPath,
+    geom = GeomFreqploy,
```
tdhock, line 19: why change this?

R/z_animint.R
```diff
@@ -959,9 +357,8 @@ animint2dir <- function(plot.list, out.dir = NULL,
       }
       geom_num <- geom_num + 1
       layer_name <- getLayerName(L, geom_num, p.name)
-      gl <- saveLayer(L, df, meta, layer_name,
+      gl <- Geom$export_animint(L, df, meta, layer_name,
```
tdhock, line 360: L$export_animint( instead?

tdhock:
looks ok, glad it passes, but there are a few unexpected changes (or are they typos?) can you please explain or fix?

R/z_animint.R
lazycipher: Yes, right! updated this!

R/geom-freqpoly.r
lazycipher: I had to introduce a new class `GeomFreqPoly` because there were certain operations which had to be done for FreqPoly geom only and that's why I had to change from `GeomPath` to `GeomFreqpoly`. The class `GeomFreqpoly` inherits `GeomPath` so everything should be good to go I believe.

R/geom-bin2d.r
lazycipher: I added this because bin2d is not supported in animint2. So, there might be chances that we'll be adding support for this. So, there should be a class where the operations can be done. And I had to show this message to users when $pre_process is being invoked.
```r
stop("bin2d is not supported in animint. Try using geom_tile() and binning the data yourself.")
```

R/geom-bin2d.r
lazycipher: I did this because I had to add a method which was supposed to be applied on `GeomBind2d` geom only.

NAMESPACE
lazycipher: Since I had to create a new class, I had to export this as well.

NAMESPACE
lazycipher: Since I had to create a new class, I had to export this as well.

R/z_animint.R
```diff
@@ -357,7 +357,7 @@ animint2dir <- function(plot.list, out.dir = NULL,
       }
       geom_num <- geom_num + 1
       layer_name <- getLayerName(L, geom_num, p.name)
-      gl <- Geom$export_animint(L, df, meta, layer_name,
+      gl <- L$geom$export_animint(L, df, meta, layer_name,
```
tdhock, line 360: is this correct? why do we have to refer to L twice? seems like we should fix that.

tdhock: or if export_animint method can't be fixed maybe keep the original saveLayer function.

lazycipher: Actually we're doing some operations/pre-processing in the export_animint() function by using L.

tdhock: looking at the original OO ggplot code it seems like the usual/consistent syntax is actually what you had originally, Geom$export_animint so I guess we should use that. sorry for the confusion.

R/geom-freqpoly.r
tdhock: that makes sense thanks

R/geom-bin2d.r
tdhock: ok that makes sense but shouldn't it be Bin2d? (not Bind2d) (no extra d before 2)

NAMESPACE
tdhock: right but shouldn't it be poly? (not ploy)

R/z_animint.R
lazycipher: So, Should I go ahead and revert the last change and use `Geom$export_animint()` instead?

R/geom-bin2d.r
lazycipher: Oh! I'm so so sorry! How could I not see that!

NAMESPACE
lazycipher: I'm so sorry! Let me correct this asap!

NAMESPACE
lazycipher: Let me fix this too!

tdhock:
yes

lazycipher:
@tdhock, I've fixed typos and changed export_animint() initiation.

tdhock:
looks good to me please merge

lazycipher:
Sure thing! Thank you so much!

tdhock:
actually on the CRAN build I see the following can you please fix?
```
* checking for missing documentation entries ... WARNING

Undocumented code objects:

  ‘GeomBin2d’ ‘GeomFreqpoly’

All user-level objects in a package should have documentation entries.

See chapter ‘Writing R documentation files’ in the ‘Writing R

Extensions’ manual.

* checking for code/documentation mismatches ... WARNING

Functions or methods with usage in documentation object 'saveLayer' but not in code:

  ‘saveLayer’
```

lazycipher:
Sure thing! Let me fix this asap!

PR #54, opened 2022-05-16
Added ability to plot rotated geom_text

csaluski:
Adds ability for geom_text to be rotated arbitrarily, closely matching the appearance of ggplot2's text rotate. The appearance is slightly different than in ggplot2 due to the fact that the rotation point in this code is the bottom center anchor of the text, while in ggplot2 it is the center of the point where the text is plotted. 
Note that this code slightly increases the rendered page output, as it adds a `transform` attribute to every text element. If we wanted to do different transforms with the `transform` attribute this could get ugly, but since translation and scale are handled with `x`, `y`, and `size` attributes, I don't think it's very likely that would be required.


tdhock:
code looks good thanks! can you please
1. try to get tests working on github actions
2. add a new test case for geom_text(angle=90), specifying angle as a geom parameter (current test only covers case of angle in aes). I believe when the geom parameter is specified, then the angle for that geom should be saved in plot.json, and then there is no need for an angle column in the tsv files.

Csaluski:
Sounds good, thanks for adding me. I tested out getting GitHub Actions set up on my fork and I think it's working. The tests don't pass but I believe that's because of regressions in the code, not an incorrect test setup. I'll make a PR from that now and then work from this repo directly in the future.

tdhock:
yes there could be regressions. which tests are failing?

Csaluski:
I think you saw them from the other PR, but for reference [renderer2 is failing 1 test](https://github.com/Csaluski/animint2/runs/6498512648?check_suite_focus=true#step:9:49), and it looks like the manual PDF for building for CRAN is hitting an error that might just be a [style file missing](https://github.com/Csaluski/animint2/runs/6498512970?check_suite_focus=true#step:9:130).

tdhock:
I have seen that inconsolata error before, but not for a long time. You need to install that latex package. If you are on ubuntu it should be a simple apt-get install fonts-inconsolata or something similar, https://packages.ubuntu.com/search?keywords=inconsolata&searchon=all&suite=all&section=all

inst/htmljs/animint.js
```diff
@@ -1087,20 +1087,28 @@ var animint = function (to_select, json_file) {
...
     const get_angle = function(d) {
+      x = scales["x"](d["x"]);
+      y = scales["y"](d["y"]);
+      // have to use optional chaining operator because params may be null 
       if (d.hasOwnProperty("angle")) {
-        x = scales["x"](d["x"]);
-        y = scales["y"](d["y"]);
-        console.log(x);
         // ggplot expects angles to be in degrees CCW, SVG uses degrees CW, so 
         // we negate the angle.
```
tdhock, line 1105: please move this comment down to where you use the negative in the code

tests/testthat/test-renderer1-text.R
```diff
@@ -42,21 +42,39 @@ plot.vec <- data.table(
...
   transform <- data.table(t(sapply(geom, xmlAttrs)))$transform
   expect_true(any(grepl("-45", transform)))
   expect_true(any(grepl("-90", transform)))
   expect_true(any(grepl("0", transform)))
+})
+
+viz.geom.angle <- list(scatter = scatter.plot <- ggplot() +
+  geom_text(
+    data=plot.vec,
+    aes(x=x, y=y, label=labs, angle=angle),
```
tdhock, line 67: shouldn't the angle be outside aes here?

Csaluski: Thanks for that catch, yes it should be.

tests/testthat/test-renderer1-text.R
```diff
@@ -30,3 +30,52 @@ test_that("text may contain commas and parentheses", {
   expect_that(any(grepl("\\(", txt)), is_true())
   expect_that(any(grepl(",", txt)), is_true())
 })
+
+### Test text rotation option
+library(data.table)
```
tdhock, line 35: can you please use data.frame instead of data.table, since we currently do not have any dependency on data.table? (this should cause the CRAN test to fail but we are not seeing it because it fails on the version string before it gets to that check)

tdhock:
looks good to me, thanks @csaluski !
Please add an entry to NEWS and then merge.

csaluski:
This is only failing because of the secret not being loaded because of cross-fork security protocols.

PR #55, opened 2022-05-19
Add github actions workflow to run tests

csaluski:
Note that it seems like tests don't pass right now, I'm unsure if this is an issue with the test configuration in GitHub Actions or if it's an issue with some regressions on the tests. 

Also note that this uses the Selenium Docker image in the steps of the workflow instead of as a service container, when I tried to use it as a service container it had issues that I believe were rooted in the fact that I couldn't attach it to the host network.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: animint2
 Title: Animated Interactive Grammar of Graphics
-Version: 2022.2.2
+Version: 2022.5.24 
 URL: https://github.com/tdhock/animint2
 BugReports: https://github.com/tdhock/animint2/issues
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


tdhock:
looks like some of the test failures are easy fixes,
```
Deprecated: please use `expect_lt()` instead
```

Csaluski:
See [test results](https://github.com/Csaluski/animint2/actions/runs/2374340656) on my fork of the repo, I think this should be ready to merge. One thing worth double checking is the change to the widerect test I made, I commented the issues I was having that lead me to change it but it would be good to figure out exactly why it wasn't working.

tdhock:
looks great +1 please update NEWS/DESCRIPTION with increased version number, then merge with master.

Csaluski:
@tdhock the tests need a Github PAT with Gist access added to the repo secrets before they'll pass, I believe you're the only person who can set this up.

Also note that the build was also failing because of a trailing space on the Version tag in DESCRIPTION, I fixed that on my branch but I'm not sure if we should make a whole new PR for that.

tdhock:
OK thanks for the info, I believe this should be fixed now on master.

PR #60, opened 2022-06-14
Add support for alpha_off 

csaluski:
Allows the user to specify an `alpha_off` parameter in the `geom` parameters or the `aes` parameters and use that opacity as the deselected opacity, instead of the previous default of `alpha - 0.5`. Replicates the default behavior if an alpha_off parameter is not provided.

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1047,6 +1047,15 @@ var animint = function (to_select, json_file) {
       }
       return a;
     };
+    const get_alpha_off = function (d) {
+      let a;
+      if (aes.hasOwnProperty("alpha_off") && d.hasOwnProperty("alpha_off")) {
+        a = d["alpha_off"];
+      } else {
+        a = base_opacity;
+      }
+      return a;
+    };
     var size = 2;
     if(g_info.geom == "text"){
       size = 12;
@@ -1570,7 +1579,9 @@ var animint = function (to_select, json_file) {
 	"opacity":{
 	  "mouseout":function (d) {
 	    var alpha_on = get_alpha(d);
-	    var alpha_off = get_alpha(d) - 0.5;
+      // 
+      let alpha_off = get_alpha_off(d);
+	    // var alpha_off = get_alpha(d) - 0.5;
 	    if(has_clickSelects){
               return ifSelectedElse(d.clickSelects, g_info.aes.clickSelects,
 				    alpha_on, alpha_off);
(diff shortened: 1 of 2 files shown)


csaluski:
Still working on tests for this, will push those soon.

csaluski:
While working on building tests for this I realized that mixing `alpha` and `alpha_off` in between the `geom` and `aes` parameters currently does not work. See [these lines](https://gist.github.com/csaluski/a465dfdd2707837c4fa8ea687156e37d).

csaluski:
There is also currently an issue where when the `alpha` is set in the `geom` parameters, mousing over a point in that geom and then removing the mouse from that point will reset the opacity of the point to `1.0` until an update is caused by changing to a different selection.

tdhock:
thanks @csaluski this looks like a good start. make sure to include comprehensive tests for the new features. I like the name alpha_off!

csaluski:
Will do, I'm working on a test suite that tests every combination of geometry, aesthetic, and unspecified `alpha` and `alpha_off`, as some of the combinations have special logic considerations. 

One thing I'm wondering is where to document this addition, as I'm not sure where the complete list of aesthetic and geometry properties is listed currently. I've found a few parameters I wasn't aware of while looking through the tests, namely `id` and `href`.

tdhock:
hi @Faye-yufan this PR is redundant with #72 right? 
So let's close this PR?

Faye-yufan:
Hi @tdhock , this PR is not redundant, actually #72 is compatible with this PR. 
Can we merge this first?

tdhock:
I'm confused, can you please clarify?
Both PRs are wanting to merge changes into master.
Your PR #72 implements both alpha_off and color_off, right?
What is the use of merging this PR then?

Faye-yufan:
Technically #72 only implements `color_off`, but I modified to let it works with the default opacity(0.5) style. I was planning to make a commit on color_off PR after this was merged, so just a small change([a line of code](https://github.com/tdhock/animint2/blob/9fbc6212112a724518022867e0c3ee89008a8b5f/inst/htmljs/animint.js#L1631-L1632)) would make both `alpha_off` and `color_off` work together. It's easy to implement alpha_off on my pr, but I think @csaluski 's tests are really great.

tdhock:
ok great that is fine to merge both then, thanks for clarifying.

Faye-yufan:
Hi @csaluski can you merge this pr? Thanks!

PR #63, opened 2022-06-20
Improve getCommonChunk() speed in compiler

### LINKED ISSUE #57: Compiler speed improvements with many selection subsets
While working on some visualizations I've run into some performance issues that seem to be related to subsetting large data sets with `showSelected` on several criteria. I've created a simplified example of the plot that I'm creating using synthetic data [here](https://gist.github.com/csaluski/91d47e262fac3e50572e6c71c5d7cd88). This code can be easily expanded to see what other variables may have an impact on compilation speed.

Profiling with profvis shows that the second system with the addition of groups takes almost twice as long to compile, although in my original code where I had performance issues it took even longer. A large majority of the time is spent in [`getCommonChunk`](https://github.com/tdhock/animint2/blob/6431aba10bc76adf5d0425a66eada21cba46f90f/R/z_animintHelpers.R#L742-L750), which I believe is called when the Geom objects are calculating how to optimize their data into chunks for file output.

It may be worth examining the performance characteristics as further subset options are added to figure out what the approximate runtime complexity is, so that future users can be aware of what performance considerations their plots have.

Faye-yufan:
### Summary
Profiling with profvis shows that [```getCommonChunk```](https://github.com/tdhock/animint2/blob/96ec0caf446c555ae847febdf3681de6a6525ee8/R/z_animintHelpers.R#L750-#L874) function takes a lot of time inside compiler, so it is worth trying to rewrite the func which was based on ```data.frame```. View the profile [here](https://bl.ocks.org/Faye-yufan/f81c1a8f27050acc3bc11b169d515c19).

```data.table```  inherits from ```data.frame``` and works perfectly even when data.frame syntax is applied on data.table. And one of its most notable features is speed. So I try to improve the speed by using data.table.

Resolve #57 

### Other concern
* Need to convert `dt` back to `df` before function return.
* Use some functions from `data.table`, the robustness would be affected?

### CODE REVIEW

**Comment by csaluski** [comment]
It looks like `data.frame` and `data.table` have different behaviors in some of these tests, I haven't deeply investigated the logs but many of them appear to be exhibiting different behaviors of merging or diffing tables, leading to errors like [this](https://github.com/tdhock/animint2/runs/6975807805?check_suite_focus=true#step:9:289).

**Comment by Faye-yufan** [comment]
Will continue testing the rest of the code, to see if the performance would be better or not.

File: R/z_animintHelpers.R [hunk 1]
```diff
@@ -760,19 +761,19 @@ getCommonChunk <- function(built, chunk.vars, aes.list){
   ## Remove columns with all NA values
   ## so that common.not.na is not empty
   ## due to the plot's alpha, stroke or other columns
-  all.nas <- sapply(built, function(x){all(is.na(x))})
-  built <- built[, !all.nas]
+  built <- as.data.table(built)
+  built <- built[,lapply(.SD, function(x) {if(all(is.na(x))) {NULL} else {x}} )]
```
tdhock, line 765: please create a new line after {

R/z_animintHelpers.R
```diff
@@ -762,7 +762,14 @@ getCommonChunk <- function(built, chunk.vars, aes.list){
...
   ## due to the plot's alpha, stroke or other columns
   built <- as.data.table(built)
-  built <- built[,lapply(.SD, function(x) {if(all(is.na(x))) {NULL} else {x}} )]
+  ## convert 'built' from df to dt to improve speed
+  built <- built[, lapply(.SD, function(x) {
+    if (all(is.na(x))) {
+      NULL
+      } else {
+        x
+        }
```
tdhock, line 771: for closing parens, please use same indentation as the line which opened them (here you have 2 and 4 too many spaces)

tdhock:
hey @Faye-yufan here is some demo code for comparing asymptotic timings of different git versions of an R package. Can you please run it on the before/after commits in this PR so we can see what the speed improvements look like for various sizes of the example that Charlie mentioned was slow?
https://github.com/tdhock/atime/blob/main/vignettes/git.Rmd

Faye-yufan:
@tdhock I modified atime code for animint in [this gist](https://gist.github.com/Faye-yufan/4e9301201a85af7c51e2b334e00eabd7), and below are the results for using dataframe and data.table.
atime-df-vs-dt-1
atime-df-vs-dt-ggplot-2

tdhock:
looks good, modest constant factor speedups, thanks for sharing!
anything else you want to do before merging?
@csaluski what do you think?

tdhock:
hi @Faye-yufan can you please merge this PR?

Faye-yufan:
Yes, thanks for the reminder!

PR #66, opened 2022-06-30
Add support for legend/axis text size

### LINKED ISSUE #64: Support for specifying legend/axis font size via theme()
Currently Animint theme() does not support arguments like [`text`](https://ggplot2.tidyverse.org/articles/extending-ggplot2.html?q=theme%20text#global-elements), [`axis.text`](https://ggplot2.tidyverse.org/reference/element.html#ref-examples) nor [`legend.text`](https://ggplot2.tidyverse.org/articles/faq-customising.html#how-can-i-change-the-font-sizes-in-the-legend) to change axis/legend text size like below:
```
+ theme(...,
        axis.text = element_text(size=X),
        legend.text = element_text(size=X),
        ...)
```
I think this feature would require add new functions in [extract theme settings](https://github.com/tdhock/animint2/blob/4b1049bc1d8ee8dbd78b31f52a5863496fae5931/R/z_animint.R#L146-L149), [add_plot func](https://github.com/tdhock/animint2/blob/92e3bc5dd6937e61c2d7bb247c37448cc9b02138/inst/htmljs/animint.js#L309-L321), [d3 axis text setting](https://github.com/tdhock/animint2/blob/92e3bc5dd6937e61c2d7bb247c37448cc9b02138/inst/htmljs/animint.js#L180) and [add_legend func](https://github.com/tdhock/animint2/blob/92e3bc5dd6937e61c2d7bb247c37448cc9b02138/inst/htmljs/animint.js#L2011-2128) in animint.js. And the axis and lengend font size maybe default to 11px and 12px seperately if user not defined.  

It may require add conditions to extract `axis.text` params or another variable name for assigning axis text size, maybe like `axis.font`. For legend text size, I think simply add `.style("font-size","XXpt")` would work.

Faye-yufan:
Enable `size` options in `axis.text.x`, ` axis.text.y`, `legend.text` to be aligned with ggplot2 syntax. Try to enable `axis.text` and `text` as well, it may require more logic in compiler.

- [x] Add relative _'size'_ parameters into the compiler for extracting input parameters in theme().
- [x] Add d3 attributes in the renderer to display the text size.
- [x] Modify plot strip/axis calculation in the renderer, to make the plot displayed properly.

closes #64


Faye-yufan:
Font size, the number, is able to be extracted from `theme()` via compiler and directly passed into JS with `px` as measurement. But don't know why ggplot2 `rel()` mechanics cannot take effect. 

Maybe it's because the [`plot_theme()`](https://github.com/tdhock/animint2/blob/ce5e887702623ee84af0f7e402f8582d28ebd9be/R/theme.r#L375-L397) function being called [inside compiler](https://github.com/tdhock/animint2/blob/ce5e887702623ee84af0f7e402f8582d28ebd9be/R/z_animint.R#L29) did not compute the relative scale size? But the no-warning compiler theme [testthat unit](https://github.com/tdhock/animint2/blob/1e7c2683edf74191a4976a46d7b334d9c3840cd5/tests/testthat/test-compiler-theme.r#L92-L98) shows that it can successfully convert the relative text size, like `rel(0.5) of 12` is 6.

Also, `rel` S3 class is numeric, `is.numeric(rel(2)) == TRUE`.

tests/testthat/test-renderer1-theme-text-size.R
```diff
@@ -0,0 +1,90 @@
...
+  viz2$one <- s2
+  info1 <- animint2HTML(viz2)
+  axis.x.style.value <-
+    getStyleValue(info1$html, '//g[@class="xaxis axis xaxis_1"]//g[@class="tick major"]//text', 
+                  "font-size")
+  axis.y.style.value <-
+    getStyleValue(info1$html, '//g[@class="yaxis axis yaxis_1"]//g[@class="tick major"]//text', 
+                  "font-size")
+  expect_match(axis.x.style.value, "20px")
+  expect_match(axis.y.style.value, "12px")
```
tdhock, line 50: these tests look good!

R/z_animintHelpers.R
```diff
@@ -162,6 +162,33 @@ setUpdateAxes <- function(theme, options_list){
...
 
 
+#' Get x/y axes font size from params text, axis.text and axis.text.x/y via theme()
+#'
+#' @param xy.size size params defined in axis.text.x or axis.text.y
+#' @param theme.pars theme settings
+#' @return size
+axesTextSize <- function(xy.size, theme.pars){
+  text.size <- if(is.null(theme.pars[["text"]]$size)){11}else{theme.pars[["text"]]$size}
+  axis.text.size <- if(is.null(theme.pars[["axis.text"]]$size)||theme.pars[["axis.text"]]$size==rel(0.8)){
```
Faye-yufan, line 172: In the output, the axis.text default to 0.8, and printed out as `0.8 *` because [this line of code](https://github.com/tdhock/animint2/blob/4b1049bc1d8ee8dbd78b31f52a5863496fae5931/R/theme-elements.r#L101) in theme-element.r

Faye-yufan: This function should be modified to be more clear and less fragile.

tdhock:
code looks good but can you please paste a screenshot of the test cases after rendering them with bigger text size? I want to make sure the web page looks reasonable / readable even after playing with the text size. (I know you have expect_ statements that check for the size of individual elements that you resized, but it is also importantant to check that the plot is still OK overall)

tdhock:
For documenting these new features probably would be good to add to Plot-specific options under Chapter 6 Animint options, https://rcdata.nau.edu/genomic-ml/animint2-manual/Ch06-other.html#plot-options

Faye-yufan:
Got it, will write the document after it merges.
This is how the axes and legend look like inside the testing environment:

tdhock:
looks great, thanks very much, please merge!

R/theme-defaults.r
```diff
@@ -98,9 +98,9 @@ theme_grey <- function(base_size = 11, base_family = "") {
     legend.key.size =    unit(1.2, "lines"),
     legend.key.height =  NULL,
     legend.key.width =   NULL,
-    legend.text =        element_text(size = rel(0.8)),
+    legend.text =        element_text(size = rel(16/11)),
     legend.text.align =  NULL,
-    legend.title =       element_text(hjust = 0),
+    legend.title =       element_text(size = rel(16/11), hjust = 0),
```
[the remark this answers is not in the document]
Faye-yufan: Hi @tdhock thanks for the info! I did not think of this situation before.
For here, I believe it's because the plot is printed by the data generated by `ggplot_build` and did not render to html, so the text size gets bigger in ggplot. 

I changed the default text sizes for the convenience of rendering html, since previously animint set legend text size to 16px and I just thought the change to `rel(16/11)` would pass 16 as default to html. 

Yes, I think we should change the defaults back in ggplot default, and change our default text size in JS side.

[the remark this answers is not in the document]
tdhock: the top image was animint/html and the bottom was static/ggplot_build/png.
that is a good idea to change back the defaults in R code (so then static/ggplot_build/png looks the same), and also to change the defaults in JS (so the animint/html is more reasonably sized). Please create another PR for that.

PR #67, opened 2022-07-01
Refactors scale-gradient.r and scale-hue.r to be much more concise, a…

csaluski:
…nd adds new color_off parameter to compiler. Not fully implemented yet, does not insert appropriate fill_off or non-aesthetic parameters.


NAMESPACE
```diff
@@ -411,6 +419,18 @@ export(scale_fill_grey)
 export(scale_fill_hue)
 export(scale_fill_identity)
 export(scale_fill_manual)
+export(scale_fill_off_continuous)
+export(scale_fill_off_date)
+export(scale_fill_off_datetime)
+export(scale_fill_off_discrete)
+export(scale_fill_off_gradient)
+export(scale_fill_off_gradient2)
```
tdhock, line 427: @csaluski I don't think this is correct... there should not be a separate scale for color_off,fill_off. The color_off and color values should be combined into a single scale.

PR #68, opened 2022-07-15
Color-off works for a single static color

Faye-yufan:
Define a single color for non-selected data. [This example](http://bl.ocks.org/Faye-yufan/raw/a03556b5c4ae261cc0a2112d1694bf73/) is defined using:
``` R
p <- ggplot() + geom_point(
  data = mtcars,
  color_off = "lightgrey",
  size = 10,
  aes(x=mpg, y=cyl,color = disp),
  clickSelects = "gear")
  
animint(p)
```


Faye-yufan:
The example is not correct for `color_off`, the example is more like `fill_off`

PR #69, opened 2022-07-19
Change ggplot legend theme back to default

Faye-yufan:
According to the [code review](https://github.com/tdhock/animint2/pull/66#discussion_r922236793) in PR #66 , when using static ggplot without html rendering, the legend text size would become too big.


Faye-yufan:
@tdhock Hi, I just changed the default size back in theme-default.r . It shall only be 16px for the legend text in animint/html. 
But I'm not sure if this can resolve the problem on the first image in the review comment, and I think I need more details on this. So is that the legend text wrapped because of the font size?
image

inst/htmljs/animint.js
```diff
@@ -2096,7 +2096,7 @@ var animint = function (to_select, json_file) {
 	        .attr("y", 14)
           .style("fill", function(d){return d["textcolour"]||1;})
 	        .style("text-anchor", "middle")
-	        .attr("font-size", function(d){return d["textsize"]||16;})
+          .attr("font-size", function(d){return d["textsize"]||1;})
```
tdhock, line 2099: please use consistent indentation

tdhock:
I don't think the text wrapping is a problem (that depends on the size of the browser window).
As long as the text size is reverted I think it should be OK.

inst/htmljs/animint.js
```diff
@@ -2096,7 +2096,7 @@ var animint = function (to_select, json_file) {
 	        .attr("y", 14)
           .style("fill", function(d){return d["textcolour"]||1;})
 	        .style("text-anchor", "middle")
-          .attr("font-size", function(d){return d["textsize"]||1;})
+          	.attr("font-size", function(d){return d["textsize"]||1;})
 	        .text("a");
```
tdhock, line 2100: indentation still does not match ...

Faye-yufan: Sorry, it must be something wrong with my vscode. I just fixed some of the other code I wrote, is that ok?

tdhock:
hi @Faye-yufan can you please merge this PR?

PR #71, opened 2022-07-29
User-configurable selection color

Faye-yufan:
Create clickSelects_color feature on the Javascript side.
Some thoughts:
- For now, it passes the color directly to JSON and is not processed by the compiler, for example, if user inputs `clickSelects_color = "violet"` in R, it will pass the _violet_ into JS instead of the color hex code _#7F00FF_.
- Change the name `clickSelects_color` to `selected.color`?


Faye-yufan:
`color_off` would be more compatible with `alpha_off` etc.

PR #72, opened 2022-08-02
Support for feature colour_off and use it with alpha_off

Faye-yufan:
The feature `colour_off`/ `color_off` enables user to specify the color/outline/stroke of the non-selected part of data. 

Here are some examples using colour_off, for [geom_line](http://bl.ocks.org/Faye-yufan/raw/e45611c3e2cd22a5be54c5e0011ee39a/) and [geom_point](http://bl.ocks.org/Faye-yufan/raw/90baf635a161b5ed84f13990d04e75c6/). For the second geom_point example, the R code would be something like this:
```R
geom_point(...
data = mtcars,
  aes(x=wt, y=mpg, fill = disp),
  clickSelects = "gear"),
  colour="red",  # selected
  colour_off="transparent",  # not selected
...)
```
where param `colour` specifies the color/outline/stroke of the selection and `colour_off` stands for the outline color of data that are not selected. 

In this PR, I also changed some functions to support using `colour_off` and `alpha_off` jointly, see the examples above,  which probably would be easier to implement further selection styles, such as `fill_off`.


Faye-yufan:
Hi @tdhock , please take a look at the examples to see if they make sense to you. 
I think the color/stroke of the legend should be set to transparent when we are using the `colour_off`, since all of them have outlines on the legend but only selected points have outlines on the plot, which is kind of confusing.

inst/htmljs/animint.js
```diff
@@ -1549,8 +1604,8 @@ var animint = function (to_select, json_file) {
             return scales.y(d["middle"]);
           })
           .style("stroke-dasharray", get_dasharray)
-          .style("stroke-width", get_size)
-          .style("stroke", get_colour);
+          .style("stroke-width", get_size);
+          select_style_fun(g_info, e);
```
tdhock, line 1608: Hi @Faye-yufan I see that `.style("stroke", get_colour)` has been replaced by `select_style_fun(g_info, e)` which I guess works but is a bit confusing to me (previously it was clear that `get_colour` was responsible for stroke, but now it is not clear that `select_style_fun` is responsible for stroke -- does it also handle the alpha/opacity?). Did you trying modifying the `get_colour` function and that did not work for some reason?

tdhock:
Those example data viz look good, totally makes sense in terms of ggplot code and output/viz you get. However I wonder about the implementation details in the JS code.

inst/htmljs/animint.js
Faye-yufan: yes, `select_style_fun` is responsible for both stroke and opacity. Since I'm still following the `ifSelectedElse` logic, which has to update selectors by get_colour/ get_colour_off two separate functions, I haven't tried modifying `get_colour` or getting rid of `ifSelectedElse`.

Faye-yufan: Will try modifying the get_colour function.

R/geom-.r
```diff
@@ -321,6 +321,15 @@ Geom <- gganimintproto("Geom",
     if(zero.size && has.no.fill){
       warning(sprintf("geom_%s with size=0 will be invisible",g$geom))
     }
+
+    ## raise warning for using _off params without clickSelects
+    has.off <- any(names(g$params) %like% "_off")
+    has.no.cs <- !any(is.cs)
+    if(has.no.cs && has.off){
+      off.vec <- grep( "_off$", names(g$params), value = TRUE)
+      warning(sprintf("geom_%s has %s which is not used because this geom has no clickSelects; please specify clickSelects or remove %s",
```
tdhock, line 330: instead of just g$geom / geom_point, can you please display g$classed or layer_name? (which should be something like geom1_point_scatter, that gives more information -- geom number and parent plot name)

Faye-yufan: Yes, on it now

tdhock:
The only renderer test with geom_tile and clickSelects is in test-renderer3-FunctionalPruning.R, and there are currently no expectations about the appearance of the geom_tile. @Faye-yufan can you please add some new expectations about what the geom_tile should look like when there is clickSelects specified but no color/color_off? Each tile should always have opacity=1 with stroke=black when selected and stroke=transparent when not selected.
Also please create a new animint in that test file, with user-specified color_off, and corresponding expectations about the stroke when selected/not.

tests/testthat/test-renderer5-ChromHMMiterations.R
```diff
@@ -67,5 +67,5 @@ test_that("fill not constant in probability legend and circles", {
 test_that("tile stroke is black", {
   stroke.vec <- getStyleValue(
     info$html, '//g[@class="geom1_tile_parameters"]//rect', "stroke")
-  expect_true(all(stroke.vec=="black"))
+  expect_color(stroke.vec, "black")
```
tdhock, line 70: hi @Faye-yufan this should fix the unexpected test failure we observed today. the text returned by getStyleValue is actually not "black" but some RGB representation, which is computed inside the new `expect_color` helper function, which you should use for your color_off tests as well.

Faye-yufan:
on it now!

tests/testthat/test-renderer5-ChromHMMiterations.R
Faye-yufan: yes, thanks so much!

tests/testthat/test-renderer2-colour.R
```diff
@@ -0,0 +1,275 @@
...
+  colour_off="grey50",
+  clickSelects = "value")
+viz.tile <- list(
+  nocol=no.col,
+  colnooff=has.col.no.off,
+  colandoff=has.col.and.off)
+
+info <- animint2HTML(viz.tile)
+
+test_that("if has clickSelects but no colour/colour_off, selection colour/stroke should be black, and transparent for not selected (no stroke)", {
```
Faye-yufan, line 197: hey @tdhock , I added some tests here including these situations on `geom_tile`:
1. clickSelects ✔️ , colour ✖️  , colour_off ✖️ : not selected = **transparent**, selected = **black**;
2. clickSelects ✔️ , colour ✔️  , colour_off ✖️ : not selected = **transparent**, selected = **colour** value;
3. clickSelects ✔️ , colour ✔️  , colour_off ✔️ : not selected = **colour_off**, selected = **colour**;

See the the [visualization here](https://bl.ocks.org/Faye-yufan/raw/4a2e093011b9213b223b2850b28a239a/), please take a look at them.

tdhock: that is a great test case. maybe for ease of visual inspection, can you please make the size bigger? I guess that is the default size=2, so maybe increase to 5 or so?

tests/testthat/test-renderer2-colour.R
```diff
@@ -0,0 +1,275 @@
...
+  clickID('no_col_2')
+  html <- getHTML()
+  stroke.col <- getStyleValue(
+  info$html, '//g[@class="geom1_tile_nocol"]//rect[@id="no_col_2"]', "stroke")
+  expect.stroke.col <- "black"
+  expect_color(stroke.col, expect.stroke.col)
+  # not selected, stroke=transparent(no stroke style)
+  node.set <- getNodeSet(info$html, '//g[@class="geom1_tile_nocol"]//rect[@id="no_col_8"]')
+  style.strs <- as.character(sapply(node.set, function(x) xmlAttrs(x)["style"]))
```
tdhock, line 206: since you use this code block twice (once here and another time in the code below) can you please make a helper function? maybe something like
```r
expect_no_style(node.set, "stroke")
```

tests/testthat/test-renderer3-ChromHMMinit.R
```diff
@@ -83,3 +83,9 @@ test_that("animation starts by default", {
   n.different <- sum(initial.fill.vec != updated.fill.vec)
   expect_gt(n.different, 0)
 })
+
+test_that("default tile colour/stroke is transparent", {
+  stroke.vec <- getStyleValue(
+    info$html, '//g[@class="geom1_tile_parameters"]//rect', "stroke")
+  expect_true(all(stroke.vec=="transparent"))
```
tdhock, line 90: please use `expect_identical(stroke.vec, rep("transparent", 9))`

tdhock:
this is a big PR, great work, just make a couple changes, edit NAMESPACE/DESCRIPTION with new version, and then please merge.

tests/testthat/test-renderer2-colour.R
Faye-yufan: yes!

tdhock:
wait! there seems to be a bug in geom_text appearance! see screenshot below,
image
the problem seems to only occur in geom_text elements which do not have clickSelects. there should be no stroke style, but there is. can you please add a test which fails for the current code, then fix?

Faye-yufan:
@tdhock thanks for pointing out the bug, will do!

Faye-yufan:
Hi @tdhock the bug should be fixed now.

tdhock:
thanks for the quick fix! This looks good to me, and the animint2 manual compiles just fine with the new code (I compiled it on my laptop) so please edit NEWS/DESCRIPTION with new version, and then merge.

PR #73, opened 2022-08-04
Fix issue geom_contour error

### LINKED ISSUE #28: geom_contour error after loading ggplot2
```r
library(animint2)
gg <- ggplot(faithfuld, aes(waiting, eruptions, z = density))+
  geom_contour()

print(gg)

loadNamespace("ggplot2")

print(gg)
```

Faye-yufan:
Replace plyr defaults() with the same function inside the package environment.
Fix #28

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -87,7 +87,8 @@ Suggests:
     mgcv,
     nlme,
     rpart,
-    svglite
+    svglite,
+    ggplot2
 Enhances: sp
 Remotes: ropensci/RSelenium@v1.7.4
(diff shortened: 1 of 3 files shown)


tdhock:
is there some test case we can add? I'm thinking some non-testthat test file, like this https://github.com/tdhock/directlabels/blob/master/tests/unattached.R

Faye-yufan:
Thanks for sharing, will do. So the non-testthat test is like simulating this specific situation?

tdhock:
great test and fix please merge!

PR #75, opened 2022-08-25
Fix data.table bug

### LINKED ISSUE #74: data table bug after compiler speed improvements
#63 started using data.table for improving the speed of the compiler, but that seems to cause an error when compiling some animints. I'm surprised we did not catch this in our test suite, why not?
```r
year.int <- as.integer(floor(time(co2)))
month.int <- as.integer(time(co2) %% 1 * 12 + 1)
year.month.chr <- sprintf("%d-%02d-15", year.int, month.int)
year.month.POSIXct <- as.POSIXct(strptime(year.month.chr, "%Y-%m-%d"))
month.str <- strftime(year.month.POSIXct, "%B")
month <- factor(month.str, month.str[1:12])
library(data.table)
co2.details <- data.table(
  year.int,
  month.int,
  month,
  year.month.POSIXct,
  ppm=as.numeric(co2))
library(animint2)
TimeSeries <- ggplot()+
  geom_line(aes(year.month.POSIXct, ppm),
            size=1,
            data=co2.details)+
  geom_point(aes(year.month.POSIXct, ppm, color=month),
             data=co2.details)+
  xlab("Time of measurement")+
  ylab("Atmospheric carbon dioxide (parts per million)")
animint(
  TimeSeries,
  first=list(month=c("May", "October")))
```
I get:
```r
> animint(
+ TimeSeries,
+ first=list(month=c("May", "October")))
Saving animint in temporary directory; specify output directory using animint(out.dir="path/to/directory")
Error in `[.data.table`(L$data, , s.name) : 
  j (the 2nd argument inside [...]) is a single symbol but column name 's.name' is not found. Perhaps you intended DT[, ..s.name]. This difference to data.frame is deliberate and explained in FAQ 1.1.
```
@Faye-yufan can you please investigate and fix?

Faye-yufan:
Fix #74 
- If using `data.table`, convert it into `data.frame` before adding showSelected for legend.
- Create a test unit for inputting data.table.


R/z_animintHelpers.R
```diff
@@ -900,11 +900,11 @@ varied.chunk <- function(dt.or.list, cols){
     dt <- dt.or.list[, cols, with=FALSE, drop=FALSE]
     u.dt <- unique(dt)
     group.counts <- u.dt[, .N, by = group]
-    if(all(group.counts$N == 1)){
-      setDF(u.dt)
+    setDF(if(all(group.counts$N == 1)){
```
tdhock, line 903: hey @Faye-yufan did you know that the if statement returns whatever is the last expression in the if or else block? so we don't need to repeat the setDF inside the if/else, we can put it outside.

tests/testthat/test-compiler-data.r
```diff
@@ -1,55 +1,49 @@
...
+                 data=co2.details)+
+      xlab("Time of measurement")+
+      ylab("Atmospheric carbon dioxide (parts per million)"))
+  expect_no_warning({
     info <- animint2dir(TimeSeries, open.browser=FALSE)
-    })
-})
\ No newline at end of file
+  })
+  expect_is(co2.details, "data.table")
```
tdhock, line 48: hey @Faye-yufan I added this expectation which ensures that the input is still a data.table after compilation.

tdhock:
hi @Faye-yufan this looks good. I made a few changes, can you please review? if you think it is ok please merge.

R/z_animintHelpers.R
Faye-yufan: That's new to me, thanks for letting me know!

NEWS
```diff
@@ -16,6 +16,9 @@ the selector called "arclength" with the value in column
 allow configurable text size via
 theme(axis.title=element_text(size=20)) etc.
 
+2022.8.25
+
+suggest data.table, use for faster compilation.
```
tdhock, line 21: hey @Faye-yufan you should increment the version and add a small comment in NEWS before merging each PR.

Faye-yufan: Got it!

R/z_animintHelpers.R
```diff
@@ -751,6 +755,8 @@ getLegend <- function(mb){
 ##' @importFrom stats na.omit
 ##' @import data.table
 getCommonChunk <- function(built, chunk.vars, aes.list){
+  group <- NULL
+  ## Above to avoid CRAN NOTE.
```
Faye-yufan, line 759: Hi @tdhock , does the CRAN NOTE mean the output from `R CMD check --as-cran`?

tdhock: right! it complains that group is not defined, so here we give a definition (even if it is not used by the code) just to silence that check NOTE.

Faye-yufan: I see, thanks!

PR #78, opened 2022-09-01
fix legend entry text size with theme grey

tdhock:
Hi @Faye-yufan I noticed an unexpected inconsistency while re-building the animint2 manual.
See screenshot below, notice how legend entry sizes are too small in the second plot. I added a test which says both should be 16px. Can you please investigate and fix?
image

### DIFF
--- a/tests/testthat/test-renderer1-theme-text-size.R
+++ b/tests/testthat/test-renderer1-theme-text-size.R
@@ -9,12 +9,21 @@ viz <- list()
 viz$scatterFacet <- scatterFacet
 info <- animint2HTML(viz)
 
+get_axes_style <- function(html){
+  axis.style.list <- list()
+  for(xy in c("x","y")){
+    xpath <- sprintf(
+      '//g[@class="%saxis axis %saxis_1"]//g[@class="tick major"]//text',
+      xy, xy)
+    axis.style.list[[xy]] <- getStyleValue(html, xpath, "font-size")
+  }
+  axis.style.list
+}
+
 # Axes text size -------------
 test_that("unspecified axis text size default to 11px", {
-  style.value <-
-    getStyleValue(info$html, '//g[@class="xaxis axis xaxis_1"]//g[@class="tick major"]//text', 
-                  "font-size")
-  expect_match(style.value, "11pt")
+  axis.style.list <- get_axes_style(info)
+  expect_match(axis.style.list$x, "11pt")
 })
 
 # rel() only works for axis.text, axis.text.x, axis.text.y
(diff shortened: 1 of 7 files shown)


tdhock:
Also I noticed the "a" character which is drawn in the tiny svg to the left of each legend entry is too small to read, currently 3.88 -- can you please add a test which fails for the current code and then fix?
image

tdhock:
Another inconsistency that is clear from this test case is that there are black axis lines drawn on the first plot but not on the second. For consistency with default ggplot2 appearance we could remove the black axis lines from the first? What do you think?

Faye-yufan:
Appreciate the feedback.
- legend size bug for `theme_grey()` and other complete themes shall be a quick fix, I hadn't thought of this before when coding #64 
- It's interesting to see the extremely small legend svg occurs, it takes the default GeomText [`default_aes = aes(size = 3.88)`](https://github.com/tdhock/animint2/blob/987cd905fbcc97f1cd7db998e81926ac57be3674/R/geom-text.r#L165). I think it should be fixed on JS side.
- Agree to remove the black xline, maybe I'll resolve it in a new PR? Seems not much related to this one.

Faye-yufan:
Thanks for the info! 
I'm trying to figure out how to distinguish if the legend text size was input or by default.

tdhock:
I don't think the logic should care whether or not the size was default. Also I think the default value of 16 should be specified in the theme, not in the getLegendList function. But we previously observed that changing the theme value to 16 results in static ggplot text which is too big.
The tricky part is that ggplot2 text size units are in points/pts (see ?gpar and ?element_text), which is not the same as animint2/svg text size units (px/pixels). 
One solution may be to to try using "pt" instead of "px" in animint JS code. "pt" does appear to be a valid CSS unit,  https://developer.mozilla.org/en-US/docs/Web/CSS/font-size#values can you please try that? In that case we probably want a default value smaller than 16 (since 16pt is probly a lot bigger than 16px).

Faye-yufan:
Interesting, thanks a lot! Will do.

Faye-yufan:
`8.8pt`(ggplot theme default) seems still smaller than `16px`(animint default) on my screen.
image

And I start considering that px is still the better option after reading this article https://www.w3.org/Style/Examples/007/units.en.html since `pt` is the absolute units, in CSS, it is not as good as `px` except for printed output, the legend entry text size might not be aligned with other elements of the plot.
> The magic unit of CSS, the px, is a often a good unit to use, especially if the style requires alignment of text to images, or simply because anything that is 1px wide or a multiple of 1px is guaranteed to look sharp.

Indeed, it looked inappropriately large on a static plot if I modified the theme. How about magnifying the font size proportionally to px, based on ggplot theme output?  something like this:
```
.style("font-size", function(d){ return d["text_size"] * 1.82 + "px"})
```

Faye-yufan:
Now the rules of legend and axis text size are:
1. Default using ggplot2 `theme_grey()` text size, with unit `pt` in CSS styles
2. If use numeric/number to define the text size, the unit would be `pt`
3. If define with a character/string, it must end with `px` or `pt`

The default is different from what animint previous looked like.
Screen Shot 2022-09-10 at 6 44 48 PM

tests/testthat/test-renderer1-theme-text-size.R
```diff
@@ -149,3 +148,9 @@ test_that("theme_grey legend entry text size is 16px", {
   expect_match(size.list$sizeNum, "16pt")
   expect_match(size.list$sizePx, "16px")
 })
+
+test_that("Warning for invalid character/string input ", {
+  viz <- list(
+    s=scatterFacet + theme(axis.text.x = element_text(size = "12p")))
+  expect_warning(animint2HTML(viz), "axis.text.x is not numeric nor character ending with \'pt\' or \'px\', will be default 11pt")
```
tdhock, line 155: this is a good test

tdhock:
This looks great, thanks for the fix. OK to merge now if you like, after you update DESCRIPTION/NEWS.
Also for a future PR: it is possible for the user to specify other text sizes in this way? (axis names, axis tick labels, facet titles, plot title)

Faye-yufan:
Yes, it is possible and easier to specify other text since we had settled down the font size strategy. Thanks!

PR #81, opened 2023-02-02
Strip and plot title text size

tdhock:
this is a follow-up to #66 and #78
I implemented custom text sizes that can be specified in R code, for plot and panel/strip titles.
Since this is very similar to your GSOC project last year @Faye-yufan can you please review?


tdhock:
also do you have any idea why the test is failing? It seems that we should be able to get the text size via text-size style, just like the axis ticks etc.

Faye-yufan:
Hi @tdhock , thanks for the follow-up!
For the test, it's probably because the `font-size` for the title is defined as a property instead of a style, and the test helper function `getStyleValue()` is for getting style only.
I assume we don't have any helper function to get the value of properties, so I created a new function called `getPropertyValue()`. It shall pass now.

R/theme-defaults.r
```diff
@@ -117,7 +117,7 @@ theme_grey <- function(base_size = 11, base_family = "") {
     panel.ontop    =     FALSE,
 
     strip.background =   element_rect(fill = "grey85", colour = NA),
-    strip.text =         element_text(colour = "grey10", size = rel(0.8)),
+    strip.text =         element_text(colour = "grey10"),
```
Faye-yufan, line 120: So we will remove the `rel` size in animint2 aesthetic?

Faye-yufan (commented):
It looks good to me!

R/theme-defaults.r
tdhock: Thanks for your review @Faye-yufan and for the fix in the tests.
I'm not sure I understand the question. I think rel is ok to use still, but this deletion of strip.text property was intended to increase the text size of the strip text, does that make sense?

Faye-yufan: Yeah sure! I get it now.

PR #82, opened 2023-03-12
Replace phantomjs with chrome as headless browser

Faye-yufan:
This pull request replaces the previous use of PhantomJS as the headless testing browser with Chrome, which was done to resolve the issue of getting Dynamic HTML via JavaScript DOM nodes. Previously, the remote driver `getPageSource()` function returned the original skeletal HTML when attempting to retrieve XML. 
Here is an example of the output returned by the PhantomJS remote driver when requesting a page:
```
> remDr$getPageSource()
[[1]]
[1] "<!DOCTYPE html><html><head>\n\n    <meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\">\n    <title>Interactive animation</title>\n    <script type=\"text/javascript\" src=\"vendor/d3.v3.js\"></script>\n    <script type=\"text/javascript\" src=\"animint.js\"></script>\n    <link rel=\"stylesheet\" type=\"text/css\" href=\"styles.css\">\n    <!-- Selectize -->\n    <script type=\"text/javascript\" src=\"vendor/jquery-1.11.3.min.js\"></script>\n    <script type=\"text/javascript\" src=\"vendor/selectize.min.js\"></script>\n    <link rel=\"stylesheet\" type=\"text/css\" href=\"vendor/selectize.css\">\n  </head>\n\n  <body>\n\n    <div id=\"plot\"> </div>\n    <script>\n      var plot = new animint(\"#plot\",\"plot.json\");\n    </script>\n\n  \n\n\n</body></html>"
```
After switching to chrome and using the headless mode with `wdman` extra arguments, most of the tests that check for html node/property values were fixed. However, there are still a few failed tests remaining.
```
> tests_init("chrome")
...
> [ FAIL 19 | WARN 1 | SKIP 5 | PASS 1825 ]
```


tdhock:
thanks, this is a great first step toward modernizing the test framework!

Faye-yufan:
I opted to use the Chromedriver in the `wdman` package instead of `Chromote` as proposed in the previous PR, since there are numerous functions in the remoteDriver test suite that are not compatible with Chromote.

However, there are still a few tests that fail in my local environment (macOS, arm64, Chrome version: 111.0.5563.110):

1. In test-renderer1-axis-angle-rotate.R, here is one of them: 
```R
Failure (test-renderer1-axis-angle-rotate.R:48:3): axis.text.x=element_text(angle=90) means transform="rotate(-90)" ──
title_box$top >= tick_box$bottom is not TRUE

`actual`:   FALSE
`expected`: TRUE 
Backtrace:
    ▆
 1. └─animint2 (local) expect_rotate_anchor(info, "-90", "end") at test-renderer1-axis-angle-rotate.R:48:2
 2.   └─testthat::expect_true(title_box$top >= tick_box$bottom) at test-renderer1-axis-angle-rotate.R:33:2
── Failure (test-renderer1-axis-angle-rotate.R:56:3): axis.text.x=element_text(angle=70) means transform="rotate(-70)" ──
title_box$top >= tick_box$bottom is not TRUE
```
Interestingly, these tests pass perfectly in Firefox. The transform="translate(x, y)" values are different between Chrome and Firefox. This could be attributed to the differences in rendering engines and the implementation of CSS styles across browsers.

2. The following error occurs in `test-shiny.R`:
```R
── Error (test-tt-inter.R:3:1): (code run outside of `test_that()`) ────────────
<packageNotFoundError/error/condition>
Error in `find.package(package, lib.loc, verbose = verbose)`: there is no package called ‘animint’
```
I suspect this is due to the old `animint` package not being supported in the current R version.

Faye-yufan:
Also some concerns:
1. Using Chrome as the headless driver for testing may result in incompatibilities, if the local Chrome browser is not updated to the available Chrome driver versions. This imposes higher requirements for the local Chrome browser version.
2. I have not yet tested this on Linux and Windows systems. Any feedback or suggestions for these operating systems would be greatly appreciated.

tdhock:
You wrote "The following error occurs in test-shiny.R" but are you sure that you were running test-shiny.R? The output says `Error (test-tt-inter.R:3:1): ` and that file, test-tt-inter.R looks old, I will remove.

tdhock:
tests_init('chrome') gives an error on windows, any idea how to fix?
```r
> tests_init("chrome")

HKEY_CURRENT_USER\Software\Google\Chrome\BLBeacon
    version    REG_SZ    111.0.5563.65

Error: C:\Users\th798\AppData\Local/binman/binman_chromedriver app directory not found.
> Serving the directory C:\Users\th798\R\animint2\tests\testthat at http://127.0.0.1:4848

> traceback()
6: stop(assertError(attr(res, "msg")))
5: assert_that(app_dir_exists(appdir))
4: app_dir(appname)
3: binman::list_versions("chromedriver") at helper-HTML.R#71
2: rev(binman::list_versions("chromedriver")[[1]]) at helper-HTML.R#71
1: tests_init("chrome")
```

Faye-yufan:
> You wrote "The following error occurs in test-shiny.R" but are you sure that you were running test-shiny.R? The output says `Error (test-tt-inter.R:3:1): ` and that file, test-tt-inter.R looks old, I will remove.

Sorry for the mix-up, the below is the error from test-shiny.R:
```R
── Error (test-shiny.R:14:1): (code run outside of `test_that()`) ──────────────
<packageNotFoundError/error/condition>
Error in `loadNamespace(x)`: there is no package called ‘animint’
Backtrace:
    ▆
 1. └─base::loadNamespace(x) at test-shiny.R:14:0
 2.   └─base::withRestarts(stop(cond), retry_loadNamespace = function() NULL)
 3.     └─base (local) withOneRestart(expr, restarts[[1L]])
 4.       └─base (local) doWithOneRestart(return(expr), restart)
```

And thanks for the feedback for the Windows environment, I'll look into that and work on a fix.

Faye-yufan:
I'm still unable to get it to work due to incompatibility between the Chrome version and the Chrome Driver version on my machine.
```
[1] "115.0.5790.102" # the Chrome version on my local machine
[1] "112.0.5615.28" "111.0.5563.64" "111.0.5563.41" "111.0.5563.19" # the available Chrome Driver versions
[5] "106.0.5249.21" "105.0.5195.52" "105.0.5195.19"
Error in tests_init("chrome") : Cannot find compatible chrome driver.
```

tdhock:
what do you think would be the way forward here? it seems like it should be possible to get a current chrome + compatible chromeDriver...

Faye-yufan:
I have utilized WebDriver managers `wdman` to get the 'latest_compatible' version of ChromeDriver, in [here](https://github.com/animint/animint2/pull/82/files#diff-d440752ad5b3ed645d8cf33aae54cc93d95f832734bedde46ce4c1942306aa27R98-R101), it should be working because I have successfully run the test before.

There are two most straightforward ways I can think of now:
1. Manually Downgrade/Upgrade Chrome: This is probably the most straightforward approach, but it’s not the most convenient one because it requires more effort to do so.
2. Wait for ChromeDriver Update, but this is a passive approach.

I think the real snag here here is the ChromeDriver has to work with the Chrome browser, even if in the headless mode, and the version of ChromeDriver typically lag behind the browser's.

tdhock:
what do they do in other projects? we are not the only ones using chromeDriver on github....

Faye-yufan:
I've looked into it and it seems that many packages like `webdriver` and `webshot` are now leaning towards using `Chromote` as their remote browser driver, maybe that's the right choice. I'm thinking of diving in and start updating our test suite functions to use `Chromote`.
Given that it'd be a significant update for our test suite, any thoughts on where to start?

tdhock:
thanks for looking at other R packages like webdriver and webshot, and finding that Chromote is now recommended over chromedriver. That is useful.
To clarify my previous comment, I meant to investigate other projects (not necessarily R packages, python/js/etc packages also maybe) which are using chrome+chromedriver/chromote for testing, and try to figure out how they handle this version incompatibility issue. My guess is that instead of always trying to use the most recent version, we should try using a somewhat recent version (say released 1 year ago) of both chrome + chromote, which work together, and should be easy to install. I guess the issue is that our requirement for testing is that we want it to work on both (1) gh actions and (2) our own personal computers, I wonder if other projects have this requirement as well?

Faye-yufan:
I see, I'll dive deeper into that. 
Just found out the latest [announcement](https://chromedriver.chromium.org/downloads/version-selection) on the chromedriver's site. Looks like Chrome and ChromeDriver are now release together. So, I'm thinking we might not face those annoying lag issues anymore.

But here's the problem: in our `tests_init()`, we use [`binman`](https://github.com/ropensci/wdman) to fetch available chromedriver versions, and identify the compatible chromedriver version based on the system's Chrome version. Use [`wdman`](https://github.com/ropensci/wdman) to start the identified chromedriver version. And use [`RSelenium`](https://github.com/ropensci/RSelenium) to interact with the started chromedriver for running tests. However, I haven't seen any updates on these packages, specially with this new ChromeDriver release adjustment.

Also, ChromeDriver used to be released over here: https://chromedriver.storage.googleapis.com/index.html. But they've changed the urls to https://googlechromelabs.github.io/chrome-for-testing/#stable. I'm thinking that this might be why the `binman` and `wdman` fail to get the latest compatible chromedriver.

tdhock:
this PR is no longer relevant because #126 implemented chromote for testing.

PR #92, opened 2023-06-03
Fix scrolling issue in Firefox

### LINKED ISSUE #90: Observations on a Scrolling Bug
## The Bug

In Firefox (but not in Edge), there's a curious scrolling bug where some webpages with some animints scroll up by themselves. It's like the webpage is possessed by an especially boring ghost.

You can experience it for yourself. Open [Chapter 7](https://rcdata.nau.edu/genomic-ml/animint2-manual/Ch07-limitations.html) in the `animint2` Manual in Firefox, scroll all the way down, and watch the webpage just slowly scroll up by itself.

Since it's browser-dependent, I'm actually not sure that this counts as a bug for `animint2` instead of, say, Firefox. But this seemed to be the best place to document it, and I wanted to write down my observations before I forget them.

## When Does the Bug Happen?

It seems to become visible when all these criteria are met:

1. The selection menu for an animint is open.
2. The selected value/s for the variable change/s while you are on the webpage.
3. Your viewport is below the selection menu (i.e. you don't see the animint or the selection menu anymore).
4. Your viewport either does not contain the selection menu or is sufficiently far away from it. In other words, the bug may not happen if the selection menu is outside your viewport, but your viewport is very close to the selection menu.
5. If the webpage has more than one animint, your viewport does not contain all selection menus that change.

## The Fix

How do you stop it from happening? Either:

1. Hide the selection menu, or
2. Pause the animation (under animation controls), or
3. Change updates (under animation controls) to some absurdly large number—that way, you finish reading the webpage before the animint updates.

## What's the Cause?

The scrolling seems to happen every time the selected value/s change. The quicker they change, the quicker the scrolling. When they stop changing, the scrolling stops, too. Furthermore, the bug doesn't show itself on all browsers.

Therefore, it seems reasonable to conclude that the scrolling bug is caused by some combination of browser contribution and either:

1. The selected values visibly changing on the webpage, or
2. Some unknown variable that causes both the visibly changing values and the scrolling bug. (For example, maybe the JavaScript code causes it to happen.)

```mermaid
graph LR;
    changing_values-->scrolling_bug;
```
```mermaid
graph LR;
    unknown_variable-->scrolling_bug;
    unknown_variable-->changing_values;
```

Faye-yufan:
fix #90 
Deferred setting of the `selector_table` height to prevent unintended scrollbar behavior. 

Used `setTimeout` with a delay of 0 ms to place the height setting operation on the JavaScript event loop, ensuring it executes after all current tasks and rendering operations are complete.

This change appears to fix an issue where the scrollbar was moving upwards when interacting with Selectize.js elements in Firefox.

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -2392,6 +2392,11 @@ var animint = function (to_select, json_file) {
       .attr("class", "table_selector_widgets")
     ;
     var selector_first_tr = selector_table.append("tr");
+    setTimeout(() => {
+      const selector_height = selector_table[0][0].clientHeight;
+      selector_table
+        .style('height', `${selector_height}px`);
+    }, 0);
     selector_first_tr
       .append("th")
       .text("Variable")


tdhock:
Code looks good to me but I wonder if it is worth it to add a test case which would fail for current master, and which would pass using this branch? Or would that be too complicated? (or not super useful?)

Faye-yufan:
I've just incorporated a test, though I expect it to fail, as I've rolled back the previous code change – a change that this test should fail.

About the test: Selenium doesn't provide a way to track the scrolling bar movement. So, I used a JavaScript command to capture the scroll bar's position.

inst/htmljs/animint.js
```diff
@@ -2392,11 +2392,11 @@ var animint = function (to_select, json_file) {
       .attr("class", "table_selector_widgets")
     ;
     var selector_first_tr = selector_table.append("tr");
-    setTimeout(() => {
-      const selector_height = selector_table[0][0].clientHeight;
-      selector_table
-        .style('height', `${selector_height}px`);
-    }, 0);
+    // setTimeout(() => {
```
tdhock, line 2395: since the fix is commented, and the tests are still passing, this seems to imply that the test is not sufficient to capture the bug, is that right? @Faye-yufan

Faye-yufan: Correct, this version of the test cannot capture the bug, I'm thinking of a new way to replicate the bug.

Faye-yufan:
I checked out the Selenium Firefox Docker image, and it looks like the bug isn't there in version 2.53.0. That's probably why the test passed even without the fix.

PR #93, opened 2023-06-08
89 sp maptools dependency

ampurr:
in progress


tdhock:
seems like this PR deletes fortify which is used in map_data, which is used several tests. I would suggest deleting this PR and trying a different approach.

ampurr:
Okay, will do! Thanks for the comment on `map_data`. As mentioned, I totally missed that in my debugging process.

PR #95, opened 2023-06-11
remove maptools dependency

### LINKED ISSUE #89: dependency on retiring spatial infrastructure packages
You will be aware, for example from:
https://r-spatial.org/r/2022/04/12/evolution.html,
https://r-spatial.org/r/2022/12/14/evolution2.html,
https://r-spatial.org/r/2023/04/10/evolution3.html and
https://rsbivand.github.io/csds_jan23/bivand_csds_ssg_230117.pdf and 
perhaps view https://www.youtube.com/watch?v=TlpjIqTPMCA&list=PLzREt6r1NenmWEidssmLm-VO_YmAh4pq9&index=1 
that `rgdal`, `rgeos` and `maptools` will be retired this
year, in October 2023.

`maptools` is suggested, and in R/fortify-spatial.r `maptools::unionSpatialPolygons` is used in line 33. Probably this needs revisiting anyway, but https://github.com/r-spatial/evolution/blob/main/pkgapi_230305_refs.csv suggests `sf::st_union` as an equivalent. The example also uses `maptools` to read a file, here `sf::st_read()` would match. It would be helpful to fix this best by June, latest October 2023.

tdhock:
Closes #89 
hi @ampurr here is a simple attempt to remove maptools dependency, what do you think?


ampurr:
@tdhock, it looks good! The only thing that gives me pause is the error message in Test Suite CRAN.

The message is weird: "Error in `library(sp)`: there is no package called 'sp'." I wonder if it's related to changes to `sp`. This error also pops up when using `build.sh`. The error appears even after ensuring that `sp` isn't somewhat downloaded in multiple directories, and it isn't fixed by uninstalling and re-installing `sp`.

rsbivand:
I don't think `sf::st_union` takes that kind of second argument, see https://r-spatial.github.io/sf/reference/geos_combine.html which was why my suggestion dropped the `region` argument. Elsewhere I've used `sf::aggregate` on `sf` objects, with `region` as the `by` argument.

tdhock:
on my computer at least, the proposed example takes more too much time so I will delete it
```
Examples with CPU (user + system) or elapsed time > 5s
            user system elapsed
facet_grid 5.393  0.027   6.879
coord_map  4.518  0.059   5.288
fortify.sp 1.618  0.102   9.627
```

DESCRIPTION
```diff
@@ -74,6 +74,7 @@ Imports:
 Suggests:
     servr,
     sf,
+    sp,
```
tdhock, line 77: @ampurr Need to list sp in Suggests in order for it to be installed and available for tests during github actions.

ampurr: @tdhock: Oh, thanks! Didn't realize.

PR #104, opened 2023-08-11
Support for feature `fill_off`

Faye-yufan:
The `fill_off` feature is an extension of the existing user-configurable selection functionality, following `color_off` and `alpha_off`(#72 ), allowing users to specify the fill of the non-selected parts of the data, enhancing the visual distinction between selected and non-selected data.

```r
geom_point(
  data = mtcars,
  aes(x=wt, y=mpg, 
         fill = disp), # selected
  clickSelects = "gear"),
  fill_off="transparent",  # not selected
)
```

This PR also includes modifications to support the combined use of `colour_off`, `alpha_off`, and `fill_off`, offering a richer set of visual cues for data selection.

**Testing:**
I'll test the new `fill_off` feature across various visualizations to ensure expected behavior, especially those geom not support fill aesthetics.


tdhock:
looks like a good start thanks!

Faye-yufan:
Hi @tdhock , this PR is ready for review, and should I go ahead and compile the manual to make sure this change doesn't mess up any existing animint visualizations? Or would it be better to write a test that compares the HTML nodes between the old and new manuals?

tests/testthat/test-renderer2-fill.R
```diff
@@ -0,0 +1,57 @@
+acontext("fill_off")
+
+#
+# test geom with both fill and colour styles
+#
+viz.point <- list(pointone = ggplot() + geom_point(
```
tdhock, line 6: can there be more informative names than pointone/two/three please? (something that indicates their purpose, default, fill_off_specified, etc)

tests/testthat/test-renderer2-fill.R
```diff
@@ -0,0 +1,57 @@
...
+  alpha_off=0.5,
+  fill_off="grey",
+  size = 10,
+  aes(x=wt, y=mpg,
+      fill = disp,
+      id=paste0("pointthree_disp", disp, "gear", gear, "wt", wt)),
+  clickSelects = "gear")+
+  ggtitle("fill_off + alpha_off"))
+
+info2 <- animint2HTML(viz.point)
```
tdhock, line 34: can you change info2 to something like viz.info (numbers can be confusing)

tests/testthat/test-renderer2-fill.R
```diff
@@ -0,0 +1,57 @@
...
+pointtwo = ggplot() + geom_point(
+  data = mtcars,
+  fill_off="transparent",
+  size = 10,
+  aes(x=wt, y=mpg,
+      fill = disp),
+  clickSelects = "gear")+
+  ggtitle("colour corresponding to `disp` group, fill_off=\"transparent\" "),
+
+pointthree = ggplot() + geom_point(
```
tdhock, line 23: seems like this test only covers geom_point. Do you think it would be worth testing another one like geom_rect? What about the geoms that are special cased in the JS code? (hline/vline/segment/linerange)

tdhock:
I suspect that the manual re-build should be fine with this branch, and I can do that to check later

tdhock:
I updated the manual to include an example of fill_off, https://github.com/tdhock/animint-book/commit/35dbca0ab7b89260418e65ab21cd5395e5b94b12

tdhock:
The other pages of the manual look fine after re-rendering everything with the new code.

tests/testthat/test-renderer2-fill.R
Faye-yufan: yes great point! I'll go ahead and rename these.

tests/testthat/test-renderer2-fill.R
Faye-yufan: Absolutely, I will add those tests. Thanks for the suggestion!

inst/htmljs/animint.js
```diff
@@ -1752,8 +1771,13 @@ var animint = function (to_select, json_file) {
 	});
       }
     }else{//has neither clickSelects nor clickSelects.variable.
-      elements.style("opacity", get_alpha)
-      if(g_info.geom != "text"){
+      elements.style("opacity", get_alpha);
+      // geom_segment/linerange/hline/vline no `stroke` with no clickSelects
+      const excludedGeoms = ["segment", "linerange", "hline", "vline"];
```
tdhock, line 1776: this logic is strange to me, why not instead enumerate the geoms that do support fill? (rather than the current code which enumerates the geoms which do not support fill)
even better, I wonder if we can define some class/inheritance structure which takes care of this kind of logic, and would avoid this kind of special case code.
GeomPoint inherits from GeomsWhichSupportFill inherits from general Geom etc?

Faye-yufan: I'm a little confused, you mean to create separate classes for each geom type, and within those classes, define the styles and behaviors that are specific to that geom? 
That we should refactor this existing code into a class-based structure?https://github.com/animint/animint2/blob/efa91c4e783569bb236e7f1702ddba7eb0946374/inst/htmljs/animint.js#L1522-L1531

tdhock: right. that would be cleaner / easier to maintain than the current approach using if/then in different places. maybe we can save that for a future PR though if you think it is too complicated?

tdhock: great

tests/testthat/test-renderer2-fill.R
```diff
@@ -1,57 +1,77 @@
...
-  expect_color(after.click.color, before.click.color)
+  expect_false(isTRUE(all.equal(before_click_fill, after_click_fill)))
+  expect_color(after_click_color, before_click_color)
 })
 
-test_that("fill and color are not same", {
-  point.xpath <- '//svg[@id="plot_pointthree"]//circle[@class="geom"]'
-  circle.list <- getNodeSet(info2$html, point.xpath)
-  circle.color <- getStyleValue(info2$html, point.xpath, "stroke")
-  circle.fill <- getStyleValue(info2$html, point.xpath, "fill")
```
tdhock: much better thanks

Faye-yufan:
Added a test for geom_rect that highlights incorrect behavior. Specifically, the stroke should remain visible when clicking on another rectangle. Below is the bug, the red rectangle should stay:

tdhock:
I'm not sure that "the stroke should remain visible when clicking on another rectangle" -- The default for rects is to use color="black" and color_off="transparent" right? 
Are you referring to this code?
```
    color = "black", fill = "blue", fill_off = "transparent", clickSelects = "category"
...
test_that("In geom_rect, fill_off only changes fill when clicked, colour does not change", {
```
in the test code above, I believe color_off is still transparent by default (since it was not specified), so the stroke should disappear when clicking another rectangle, right?

Faye-yufan:
Sorry about the confusion, I've updated the code in the previous comment. And yes, I was referring to that code, the only difference from the test code is that I changed the values for `color` and `fill` to better illustrate their differences. 
Since I didn't set a `color_off` parameter, I think it should not be 'triggered' and the stroke color should remain the same when clicked?

tdhock:
"Since I didn't set a color_off parameter, I think it should not be 'triggered' and the stroke color should remain the same when clicked" -> You are right that you did not set a color_off parameter, but the default color_off for rects is transparent, so if you want it to remain the same when clicked, you should specify color_off=same as color, right?

Faye-yufan:
Oh yes you are right! We do have these tests for the geom_rect/tile. Paste the code here just for my own information: https://github.com/animint/animint2/blob/12b57c9c03969cfdbdc41690162329102905fd48/tests/testthat/test-renderer2-colour.R#L212-L239

Faye-yufan:
Hi @tdhock , can we merge this pr before refactoring the geoms?
For now the `fill_off` test covers geom_point, geom_rect and geom_vline, are there any other edge cases or specific scenarios you think we should include in the tests?

tdhock:
sure, looks good to me, please update NEWS.md and then merge.

Issue #48, opened 2020-08-20
Move data subset calculation from compiler to the renderer.


lazycipher:
Currently, The computations of updating of axes/legends after changing the currently displayed data subset are done in the compiler but there are some limitations, so it will be good to move the computations to the renderer. 
We sometimes only have a subset of the whole data while computing the scales in the compiler so we are limited in our implementation. 
This can be solved by moving the computation to the renderer so that we have the whole data to compute new scale values.

Faye-yufan:
Hi @tdhock @faizan-khan-iit , I started to look at the code for computing the domain from last week, As I understand it, the logic for `update_axes` on the compiler side can be found in these two files: [z_animint.R ](https://github.com/animint/animint2/blob/12b57c9c03969cfdbdc41690162329102905fd48/R/z_animint.R#L433-L538), [z_animintHelper.R](https://github.com/animint/animint2/blob/12b57c9c03969cfdbdc41690162329102905fd48/R/z_animintHelpers.R#L397-L556). 
- The `compute_domains` function seems to compute the scale range for each geom. This function returns the computed domain (scale range) for the specific geom and data subset.
- The `get_domain`  function takes in a list of subset domains(from the above function) and computes the final domain. The final domain is computed such that it contains all the subset domains.

On the renderer's end (within the JS code) at [update_scales()](https://github.com/animint/animint2/blob/12b57c9c03969cfdbdc41690162329102905fd48/inst/htmljs/animint.js#L1870C1-L2007C1), the renderer would receive the subset data(the .tsv files) from the compiler, with other associated meta-data, like the saved ranges from plot.json.

I read through the original PR you mentioned above, it seems you're suggesting shifting the `compute_domains` and `get_domain` functions to the JS side. Is that correct?

tdhock:
right!

Faye-yufan:
I'm looking into the [for loop here](https://github.com/animint/animint2/blob/12b57c9c03969cfdbdc41690162329102905fd48/R/z_animint.R#L434-L539) to see if we can move all of its variables over to the JS side. I have some problem coming up with the implementation because it looks like not all ggplot info in the environment is making its way to plot.json.

For example, `ggplot.list` doesn't seem to be directly exported. This causes issues when trying to assign values in JS, like setting `built_data` based on this layers$data: https://github.com/animint/animint2/blob/12b57c9c03969cfdbdc41690162329102905fd48/R/z_animint.R#L493-L494, which is an input parameter for `compute_domains`.

[...]

PR #107, opened 2023-10-09
Refactor `draw_geom()` and turn into a Geom class for scalability

Faye-yufan:
### Overview
This PR aims to refactor the existing `draw_geom` function, turning some procedure and functional code into class methods. The changes improve the code scalability, particularly when we want to implement new features on the renderer side. 
This PR is critical for future developments of issue #48 

### Changes
- Refactored the Geom class to better adhere to OOP principles.
- Added methods for selection and styling, making it easier to apply custom features.
- Optimized the logic for preparing data.

### TODO
- [ ] Create group/ungroup virtual classes inherits from the basic geom class.
- [ ] Create specific geom class inherits from the virtual classes.
- [ ] Optimize the `selected_funs`, as we discussed previously.
- [ ] Delete the old code and test complete with the new code.
- [ ] Document function logic within code


tdhock:
looks like a good start, thanks for sharing.

Faye-yufan:
Hi @tdhock , I am trying to understand the logic behind `key_fun` and `id_fun`. What are their functionality? I noticed they have been defined twice, the first time is here https://github.com/animint/animint2/blob/3b8dc699cdb1a5247151edb9d83161351b979a92/inst/htmljs/animint.js#L1224-L1232
The second time, they are part of an if-else clause, where their behavior changes when the geom type is one of the following: line, path, ribbon, or polygon.
https://github.com/animint/animint2/blob/3b8dc699cdb1a5247151edb9d83161351b979a92/inst/htmljs/animint.js#L1340-L1348

I've experimented with an animint viz using the `key` aes and attempted debugging through the web developer tools. However, I couldn't seem to trigger this function. Following the comment in the code, it appears to be related to group the data into 1 path.
https://github.com/animint/animint2/blob/3b8dc699cdb1a5247151edb9d83161351b979a92/inst/htmljs/animint.js#L1248-L1252 

Do you recall the logic behind this? Thanks.

tdhock:
Hi Yufan, thanks for asking about key vs id. 
The key_fun in JS is supposed to lookup the value which was specified in aes(key). key_fun is used in the context of a D3 data-bind, in JS code that is `elements.data(kv, key_fun)` which binds the data in kv to the svg elements, matching them according to keys from key_fun. This feature makes smooth transitions possible. (duration option in animint)

id_fun is used to assign the HTML id property using this JS code, `elements.attr("id", id_fun)` -- again the id_fun is supposed to lookup whatever value was specified as aes(id). This feature is used extensively in testing, where we lookup rendered HTML elements by id, and check their computed properties against expected values.

the reason why they are defined twice, is becuase there are two cases, based on whether the geom is grouped (path, ribbon, or polygon) or not (others). The first definition `var id_fun =function(d){   return d.id; }` is for non-grouped geoms, and the second definition `key_fun = function(group_info){  return group_info.value; }` is for grouped geoms. It would be great if you could please create virtual classes GroupedGeom and UngroupedGeom to better organize this logic. 
It is necessary to treat them separately because for un-grouped geoms, there is one HTML element to create for every data point/csv row. But for grouped geoms, there is one HTML element to create for every aes(group) value (several data points/csv rows per group).
Does that help?
btw this should be documented somewhere, where do you think would be the best place? Maybe a wiki page about "JS internals" ?

Faye-yufan:
Thanks a lot for the detailed explanation! It is definitely helpful for me, and I love the idea of creating virtual classes for group and ungroup geoms.
As for documenting this, I can put the logic within the code as comment, and once the refactoring is complete, I can then consolidate the logic of functions into a wiki page.

Faye-yufan:
If I understand correctly, **group** geoms are those `g_info.data_is_object == true`,  which includes **line**, **path**, **ribbon** and **polygon**.
Whereas **ungrouped** geoms are those defining styles in if clause - **segment**, **linerange**, **vline**, **hline**, **text**, **point**, **tallrect**, **widerect**, **rect**, **boxplot**.

tdhock:
hi @Faye-yufan when you get some time, can you please try to merge new master into this branch and resolve conflicts? There are some substantial new changes which I merged into master recently, which remove some duplicated logic in animint.js and so would be complementary to the OO/class re-organization you have been working on in this branch.

Faye-yufan:
@tdhock Thank you for modifying the logic, it helps a lot! I'm working on it now.

Faye-yufan:
Here are some notes for myself, please correct me if I'm wrong - 
1. This part of code is for process data grouping and subset selection, it also check if any '.variable' aes attribute exist(I guess this is from the original animint package before animint2?). The loop gets selected variables into an 2D array: https://github.com/animint/animint2/blob/fb8f29760c605b8ffaa26e4884572e13cf1b1d5b/inst/htmljs/animint.js#L950-L990
A question -  Does this part ever going to be used in the code? Because the `selected_array` just got initialized above and there is nothing in it. If we do the loop right after that, it would not affect anything.
https://github.com/animint/animint2/blob/fb8f29760c605b8ffaa26e4884572e13cf1b1d5b/inst/htmljs/animint.js#L954-L969
<br/>
<br/>

2. This part is for getting the current one row of data, convert the data type from `object` into `array`:
https://github.com/animint/animint2/blob/fb8f29760c605b8ffaa26e4884572e13cf1b1d5b/inst/htmljs/animint.js#L1000-L1024

tdhock:
yes the .variable suffix comes from animint (not 2) which required users to specify `aes(clickSelects.variable=selector_name, clickSelects.value=selector_value)` etc. In animint2 user specifies `geom(clickSelects=c("selector_name"=selector_value))` but under the hood (not user visible) we still use the `aes(clickSelects.variable)` etc.

I believe the `selected_arrays` parts of the code are indeed used, for selecting the subset of data which corresponds to the current showSelected selection, to display.

Faye-yufan:
> I looked at the `renderer-geom-refactor` branch and found that the class work (GeomPoint, GeomSegment etc.) was lost during the merge commit `079df5be`. The classes exist in earlier commits like `4ea7bc0e` but are not in the current branch HEAD. 

Hi @AviraL0013 , the updated (at that time) master branch was merged into this draft branch on that commit. I believe the master branch has been updated a  lot since then, please refactor based on the latest master branch.

AviraL0013:
Hi @Faye-yufan and @tdhock,
Thank you for the clarification! Based on your advice, I'll start fresh from the latest master branch rather than trying to restore from the renderer-geom-refactor HEAD. It would be really helpful if you could mentor me through this work , I'll open a draft PR early and ask for reviews as I go.

PR #112, opened 2023-12-14
fix space free vertical

tdhock:
problem to fix is shown in new test case
image
space=free means that both horizontal and vertical panels should have same distance between axis tick marks, but that is only true for x axis above (not y axis). see below for static rendering to see how it is supposed to look like
image

### DIFF
--- a/tests/testthat/test-renderer1-facet-space.R
+++ b/tests/testthat/test-renderer1-facet-space.R
@@ -12,7 +12,13 @@ viz <- animint(
     facet_grid(.~am, scales="free", labeller=label_both),
   fixed = no.panels +
     ggtitle("space=fixed, scales=fixed")+
-    facet_grid(.~am, labeller=label_both))
+    facet_grid(.~am, labeller=label_both),
+  freeBothFlip = ggplot(mtcars, aes(wt, mpg)) + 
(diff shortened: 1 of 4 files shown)


tdhock:
looks like the issue is in the compiler,
```r
> info=animint2dir(viz)
> info$plots$freeBothFlip$layout$width_proportion
[1] 0.6277958 0.3722042 0.6277958 0.3722042
> info$plots$freeBothFlip$layout$height_proportion
[1] 0.5 0.5 0.5 0.5
```

tests/testthat/test-renderer1-facet-space.R
```diff
@@ -99,6 +99,29 @@ test_that("width_proportion is constant or variable", {
   expect_true(!both.equal(info$plots$freeBoth$layout$width_proportion))
 })
 
+viz <- animint(
+  freeBothFlip = ggplot(mtcars, aes(wt, mpg)) + 
+    ggtitle("space=free, scales=free") +
+    scale_x_continuous(breaks=seq(1, 9)) +
```
tdhock, line 105: why is this line here?

tdhock: @lamtung16 please review

PR #114, opened 2023-12-17
Create test-renderer3-WorldBankPath.R

siddhesh195:
Renderer test for World bank plot of last 5 years


tdhock:
thanks for the PR. the test looks reasonable, but one thing that is potentially confusing is the different number of expected elements for circle/path. Can you please propose changes to clarify why they should be different?

siddhesh195:
The expected number of circle/path is different because of missing points in WorldBank data set. Added a comment to the test file explaining the same

PR #117, opened 2024-02-05
Args validation checks

siddhesh195:
Follow up to PR116

Added my name as contributor in DESCRIPTION
Increased version
Added a NEWS item
Added a fix for empty animint() and a test. I missed to commit the test first and will do it from the next time.

### DIFF
--- a/R/z_print.R
+++ b/R/z_print.R
@@ -85,7 +85,7 @@ animint <- function(...){
   
   # Check if argument list is empty
   if(length(L) == 0) {
-    stop("No arguments passed to animint. Arguments should 1 or more include ggplot and 0 or more options")
+    stop("No arguments passed to animint. Arguments should include ggplots(1 or more) and options(0 or more)")
   }
   
   # Generate a list of duplicate named arguments
--- a/R/z_print.R
+++ b/R/z_print.R
@@ -83,6 +83,11 @@ print.animint <- function(x, ...){
 animint <- function(...){
   L <- list(...)
   
+  # Check if argument list is empty
+  if(length(L) == 0) {
+    stop(paste("No arguments passed to animint"))
+  }
+  
   # Generate a list of duplicate named arguments
   duplicate_args <- names(L)[duplicated(names(L))]
   
--- a/R/z_print.R
+++ b/R/z_print.R
@@ -87,7 +87,7 @@ animint <- function(...){
   duplicate_args <- names(L)[duplicated(names(L))]
   
   if (length(duplicate_args) > 0) {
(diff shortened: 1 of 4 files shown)


R/z_print.R
```diff
@@ -83,6 +83,11 @@ print.animint <- function(x, ...){
 animint <- function(...){
   L <- list(...)
   
+  # Check if argument list is empty
+  if(length(L) == 0) {
+    stop(paste("No arguments passed to animint"))
```
tdhock, line 88: Could you please make the error message more informative? Please tell the user how to fix the problem.
something like "No arguments passed to animint, but arguments should be ggplots (1 or more) and options (0 or more)"
Also please remove paste as it is not necessary.

tests/testthat/test-compiler-errors.R
```diff
@@ -192,3 +192,10 @@ test_that("animint(out.dir = 'dir1', out.dir = 'dir2') is an error", {
     animint(out.dir = 'dir1', out.dir = 'dir2')
   }, "Duplicate arguments are passed to animint. Duplicate arguments found: out.dir")
 })
+
+test_that("animint() is an error", {
+
```
tdhock, line 197: please remove empty line inside test_that block.
empty lines like this make it more difficult to run the whole block of code interactively. (that can be done with C-c C-c in emacs, etc)

tdhock (changes requested):
good start, please address comments and then I will merge.

tests/testthat/test-compiler-errors.R
```diff
@@ -195,5 +195,5 @@ test_that("animint(out.dir = 'dir1', out.dir = 'dir2') is an error", {
 test_that("animint() is an error", {
   expect_error({
     animint()
-  }, "No arguments passed to animint. Arguments should include ggplot(1 or more) and options(0 or more)")
```
tdhock: the old version is more correct english (although I would say ggplots instead of ggplot), can you please revert?
why did you make this change?
did you know about expect_error(something, "message", fixed=TRUE) ?

tdhock (changes requested):

tests/testthat/test-compiler-errors.R
siddhesh195: Reverted the error message to the old one and used fixed=TRUE. I made this change because I thought it would be better to avoid '()' in the error message sentence. I had used escape qualifiers to do the pattern matching in the test and didn't know about fixed=TRUE !

tdhock:
looks great thanks!

PR #119, opened 2024-03-07
Throw error when same argument is passed to aes and geom

### LINKED ISSUE #110: error in compiler when aes used at same time as param
for example, using geom(aes(alpha), alpha_off or alpha = 0.4) should be an error.

siddhesh195:
Throw error when same argument is passed to aes and geom
Closes #110 

Because of the change,  the test "empty data overrides plot defaults" would always error out since it aes(mpg, wt) is passed along with x,y to plot


tdhock (approved):
great thanks!

tdhock:
Can you please add a test for geom(aes(alpha),alpha_off) ?

siddhesh195:
> Can you please add a test for geom(aes(alpha),alpha_off) ?

Added test for geom(aes(alpha),alpha_off). This required additional changes to match _off parameters.

R/geom-.r
```diff
@@ -147,6 +147,17 @@ Geom <- gganimintproto("Geom",
...
   },
+  
+  get_off_params = function(input_string) {
+    pattern <- "_off$"
+    # Return parameter name without the _off suffix
+    if (grepl(pattern, input_string)) {
+      result <- sub(pattern, "", input_string)
+      return(result)
+    } else {
+      return("")
```
tdhock, line 158: it is simpler to not use return
also this whole function could be replaced by grep(input_string, pattern, value=TRUE)

siddhesh195: Thanks for the suggestion. I rewrote the check using grep and sub and the whole function is not required anymore

tests/testthat/test-compiler-errors.R
```diff
@@ -197,3 +196,33 @@ test_that("animint() is an error", {
...
+    y=fertility.rate, 
+    color=region,
+    group=country, alpha=population),alpha_off=0.5,
+    data=WorldBank)
+  viz <- list(
+    plot = scatter
+  )
+  expect_error({
+    animint2dir(viz, open.browser=FALSE)
+  }, "Same argument cannot be passed to both aes and geom. Argument passed to both aes and geom: alpha")
```
tdhock, line 227: can you please change the error message? it is not true that alpha was passed to both aes and geom, because alpha_off was passed to geom.
Also it would be good to explain why the same argument cannot be used / how to fix: the visual property needs only be defined in one place, so if the alpha should be different for each rendered geom, but not depend on selection state, then use aes(alpha); but if the alpha should depend on the selection state then use geom(alpha,alpha_off).
and actually for this test to make more sense with alpha_off, you would need to add clickSelects="some_selection_variable" so that the alpha_off could actually be used.
In fact could you please add another test for an error message when alpha_off is used in a geom without clickSelects? that should be a different error (alpha_off defines the opacity of geoms with clickSelects, but are not currently selected, so please fix by removing alpha_off or adding clickSelects)

siddhesh195: I updated the error message to distinguish between alpha and alpha_off. Is the new error message better to understand for the user ?

Added a new test without clickSelects and alpha_off and verified that the correct warning is displayed. When aplha_off is used without clickSelects, currently a warning is thrown. Does it make more sense to throw an error instead ?

tests/testthat/test-compiler-errors.R
```diff
@@ -209,20 +209,35 @@ test_that("Same argument passed to aes and geom is an error", {
...
-    group=country, alpha=population),alpha_off=0.5,
+    group=country, alpha=population),clickSelects="country", alpha_off=0.5,
     data=WorldBank)
   viz <- list(
     plot = scatter
   )
   expect_error({
     animint2dir(viz, open.browser=FALSE)
-  }, "Same argument cannot be passed to both aes and geom. Argument passed to both aes and geom: alpha")
+  }, "Same argument cannot be passed to both aes and geom. Argument passed to aes: alpha . Arguments passed to geom: alpha_off . The visual property needs only be defined in one place, so if it should be different for each rendered geom, but not depend on selection state, then it should be defined in aes; but if the property should depend on the selection state then it should be defined in geom")
```
tdhock, line 227: this is better but still "argument" may be confusing, maybe change to "same visual property cannot be defined in both aes and geom" ?
Also please remove space before period: "alpha ." -> "alpha." try paste0 instead of paste.

tests/testthat/test-compiler-errors.R
```diff
@@ -209,20 +209,35 @@ test_that("Same argument passed to aes and geom is an error", {
...
+    y=fertility.rate, 
+    color=region,
+    group=country),alpha_off=0.5,
+    data=WorldBank)
+  viz <- list(
+    plot = scatter
+  )
+  expect_warning({
+    animint2dir(viz, open.browser=FALSE)
+  }, "geom1_path_plot has alpha_off which is not used because this geom has no clickSelects; please specify clickSelects or remove alpha_off")
```
tdhock, line 242: great!

tests/testthat/test-compiler-errors.R
siddhesh195: Thanks, I rewrote the error message with the updated text and removed space before period

tests/testthat/test-compiler-errors.R
siddhesh195: Currently if there are multiple errors, one error masks the others since the compiler stops upon an error. Maybe this can be modified to throw multiple errors before stopping when its possible to continue compilation. gcc/llvm do that and continue compilation to discover additional errors

tests/testthat/test-compiler-errors.R
tdhock: great

tdhock:
thanks very much for this useful contribution

PR #126, opened 2024-06-06
Add chromote as headless browser for testing

siddhesh195:
Use chromote as a headless browser for running tests and migrate tests to work with chromote. 

Closes #124 

Closes #125

PR #101, opened 2023-07-13
created animint2pages()

Faye-yufan:
A function to put animint object on to github pages

PR #138, opened 2024-06-28
Set `gh-pages` as default branch of new repo created by animint2pages

### LINKED ISSUE #134: set default branch of new repo in animint2pages
hi @Faye-yufan after using animint2pages for a while I noticed that there is a small usability concern.
the default branch in new data viz repos is "main" but that is not actually used for anything.
only the "gh-pages" branch is used.
So it would be good if we could designate "gh-pages" as the default branch, which is possible on the command line via
```
git branch -M gh-pages
```
is it possible in gert / animint2pages?
That would make it easier for users, which have main as the default branch when they clone (but gh-pages as default would be more user-friendly, so the user would not have to run git branch gh-pages manually).

Faye-yufan:
Resolve #134 
After push animint files onto `gh-pages` branch, change the viz repo default branch to `gh-pages`.

### DIFF
--- a/R/z_pages.R
+++ b/R/z_pages.R
@@ -89,7 +89,7 @@ animint2pages <- function(plot.list, github_repo, commit_message = "Commit from 
     initial_commit(local_clone, repo, viz_url)
   }
   # Handle gh-pages branch
-  manage_gh_pages(repo, to_post, local_clone, commit_message)
+  manage_gh_pages(repo, owner, to_post, local_clone, commit_message, github_repo)
   message(sprintf(
     "Visualization will be available at %s\nDeployment via GitHub Pages may take a few minutes...", viz_url))
   viz_owner_repo
@@ -119,7 +119,7 @@ initial_commit <- function(local_clone, repo, viz_url) {
   gert::git_push(repo = repo, remote = "origin", set_upstream = TRUE)
 }
 
-manage_gh_pages <- function(repo, to_post, local_clone, commit_message) {
+manage_gh_pages <- function(repo, owner, to_post, local_clone, commit_message, gh_repo_name) {
   branches <- gert::git_branch_list(local = TRUE, repo = repo)
   if (!"gh-pages" %in% branches$name) {
     gert::git_branch_create(repo = repo, branch = "gh-pages")
@@ -129,6 +129,13 @@ manage_gh_pages <- function(repo, to_post, local_clone, commit_message) {
   gert::git_add(files = ".", repo = repo)
   gert::git_commit(message = commit_message, repo = repo)
   gert::git_push(remote = "origin", set_upstream = TRUE, repo = repo, force = TRUE)
+  # Set 'gh-pages' as the default branch on GitHub using GitHub API
+  gh::gh(
+    "PATCH /repos/{owner}/{repo}",
+    owner = owner,
+    repo = gh_repo_name,
+    default_branch = "gh-pages"
+  )
 }
 
 check_no_github_repo <- function(owner, repo) {
--- a/R/z_pages.R
+++ b/R/z_pages.R
@@ -31,7 +31,7 @@
 #'   viz,
 #'   github_repo = "my_animint2_plots",
 #'   commit_message = "New animint",
-#'   private = TRUE)
+#'   private = FALSE)
 #' }
 #'
 #' @export
@@ -138,10 +138,10 @@ manage_gh_pages <- function(repo, owner, to_post, local_clone, commit_message, g
   )
 }
 
-check_no_github_repo <- function(owner, repo) {
+check_no_github_repo <- function(owner, gh_repo) {
   tryCatch(
     {
-      gh::gh("/repos/{owner}/{repo}", owner = owner, repo = repo)
+      gh::gh("/repos/{owner}/{repo}", owner = owner, repo = gh_repo)
       TRUE
     },
     "http_error_404" = function(err) FALSE


Faye-yufan:
The tests for animint2pages all passed on my local. 
But in GH action, the 'PATCH' command seems to be failing due to the token in the environment.

tests/testthat/test-compiler-ghpages.R
```diff
@@ -45,3 +45,17 @@ test_that("animint2pages raises an error if no GitHub token is present", {
...
+test_that("animint2pages() default branch is gh-pages", {
+  whoami <- suppressMessages(gh::gh_whoami())
+  owner <- whoami[["login"]]
+  test_default_branch <- function(owner, repo_name) {
+    repo_info <- gh::gh(
+      "GET /repos/:owner/:repo",
+      owner = owner,
+      repo = repo_name
+    )
+    repo_info$default_branch == "gh-pages"
```
tdhock, line 58: this looks like a good start. Thanks.
however, I think a better test would be if you clone the repo and test if branch name of the clone is gh-pages ... That test would more directly address the usability concern (user would not have to switch branches) can you please change?

tdhock (changes requested):

tests/testthat/test-compiler-ghpages.R
Faye-yufan: Great idea! I will do that.

tests/testthat/test-compiler-ghpages.R
```diff
@@ -49,13 +49,15 @@ test_that("animint2pages raises an error if no GitHub token is present", {
...
-    repo_info$default_branch == "gh-pages"
+  local_repo_path <- tempfile(pattern = "repo_clone_")
+  gert::git_clone(url = paste0("https://github.com/", owner, "/animint2pages_test_repo.git"), path = local_repo_path)
+  # Check the default branch after clone
+  head_file <- file.path(local_repo_path, ".git", "HEAD")
+  head_content <- readLines(head_file)
+  if (startsWith(head_content, "ref: refs/heads/")) {
+  default_branch <- sub("ref: refs/heads/", "", head_content)
+  } else {
+  default_branch <- NA
```
tdhock, line 60: please use consistent indentation
and please use the assignment before if in cases like this
```r
default_branch <- if( some condition ) {
  some_value
} else {
  other_value
}
```

tdhock (changes requested):

tests/testthat/test-compiler-ghpages.R
```diff
@@ -45,3 +45,19 @@ test_that("animint2pages raises an error if no GitHub token is present", {
...
   file.copy(config.old, config.file, overwrite = TRUE)
 })
+
+test_that("animint2pages() default branch is gh-pages", {
+  whoami <- suppressMessages(gh::gh_whoami())
+  owner <- whoami[["login"]]
+  local_repo_path <- tempfile(pattern = "repo_clone_")
+  gert::git_clone(url = paste0("https://github.com/", owner, "/animint2pages_test_repo.git"), path = local_repo_path)
+  # Check the default branch after clone
+  head_file <- file.path(local_repo_path, ".git", "HEAD")
```
tdhock, line 55: instead of looking at the .git/HEAD file, can you please use gert as below? should be less testing code and probably more robust
```r
> gert::git_branch("~/R/animint2")
[1] "chromote-testing"
```

tdhock (changes requested):

tdhock:
the error that we are getting on github actions is below, do you know why? how to fix?
```
══ Failed tests ════════════════════════════════════════════════════════════════
── Error ('test-compiler-ghpages.R:26:3'): animint2pages() returns owner/repo string ──
<github_error/http_error_403/rlang_error/error/condition>
Error in `gh::gh("PATCH /repos/{owner}/{repo}", owner = owner, repo = gh_repo_name, 
    default_branch = "gh-pages")`: GitHub API error (403): Resource not accessible by personal access token
ℹ Read more at <https://docs.github.com/rest/repos/repos#update-a-repository>
Backtrace:
    ▆
 1. └─animint2::animint2pages(viz, github_repo = "animint2pages_test_repo") at test-compiler-ghpages.R:26:3
 2.   └─animint2:::manage_gh_pages(...)
 3.     └─gh::gh(...)
 4.       └─gh:::gh_make_request(req)
 5.         └─gh:::gh_error(resp, gh_req = x, error_call = error_call)
 6.           └─cli::cli_abort(...)
 7.             └─rlang::abort(...)
── Error ('test-compiler-ghpages.R:[53](https://github.com/animint/animint2/actions/runs/9816550931/job/27106887489?pr=138#step:11:54):3'): animint2pages() default branch is gh-pages ──
<GIT_ERROR/libgit2_error/error/condition>
Error in `libgit2::git_clone`: unexpected http status code: 404
Backtrace:
    ▆
 1. ├─gert::git_clone(...) at test-compiler-ghpages.R:53:3
 2. └─gert:::raise_libgit2_error(...)

[ FAIL 2 | WARN 0 | SKIP 0 | PASS 7[59](https://github.com/animint/animint2/actions/runs/9816550931/job/27106887489?pr=138#step:11:60) ]
```

ampurr:
Currently busy for the next few days, but on an initial glance it looks like it may be an access token problem. In that case someone will need to remake it. I wrote briefly about the procedure to do that in our internal documentation about the website:

https://github.com/animint/animint2/blob/master/readme_website.md#refreshing-the-github-access-token

The token expires every so often. GitHub says it’s a security thing. I am not qualified to say whether this is actually an effective security practice or not.

Additional documentation that could be useful:

https://usethis.r-lib.org/articles/git-credentials.html

tdhock:
the error does not say anything about expired token.
It says 403 (permission denied) and 404 (not found).

Faye-yufan:
### The first failed test:
The first test `animint2pages() returns owner/repo string` is for pushing a viz repo from local to GitHub remote. I didn't change any code in this test, so it remains the same as in the recent PR #126, where it did not fail (see [#126 recent run happened after this PR failed run](https://github.com/animint/animint2/actions/runs/9824372566)).

Therefore, the failure in the current GH Actions test is likely due to the newly added code for setting the default branch.

I believe the issue might be caused by the newly added code (https://github.com/animint/animint2/pull/138/files#diff-6c77230e5f2bdcb3a3c21e98224ada8f337bb04f677ede3b90406a3933087509R132-R138), which directly uses the GitHub API `gh`.

I am trying to use `gert` instead of `gh` to see if it resolves the issue.

### The second failed test:
It is probably because the first test failed, so no repo can be fetched through `gert::git_clone`.

ampurr:
Oh, okay. The error mentioned "Resource not accessible by personal access token" and I made a quick assumption. If it's nothing to do with the PAT, that's fine, though it also means that the solution isn't gonna be as simple as refreshing the token. Which is too bad.

Faye-yufan:
The `animint2pages() returns owner/repo string` test issue, 403 (permission denied), now has been resolved. If user want to create a new viz repo, in the initial commit, change the name of master branch to gh-pages, so it is now the default and only branch in the new repo.

The second issue, 404 (not found), remains. Still looking into it.

Faye-yufan:
I see the test repo is created under @tdhock 's personal GH, I suppose it is because of this setting?
https://github.com/animint/animint2/blob/be99e4582054c8f2655db640535f1e320878764e/.github/workflows/tests.yaml#L41-L42

But in above test run, I tried to print the output of `whoami[["login"]]` in the gh actions enviroment and the log showed
```
[1] "GitHub Pages"
NULL
```
Indicates that the gh action is trying to clone "https://github.com/NULL/animint2pages_test_repo.git", which obviously does not exist.

Let me try using credential.

tests/testthat/test-compiler-ghpages.R
Faye-yufan: Updated, thank you!

tests/testthat/test-compiler-ghpages.R
Faye-yufan: Done.

Faye-yufan:
Hi @tdhock , do you recall how you created this test repo https://github.com/tdhock/animint2pages_test_repo ? Was it pushed from your local machine or via a GitHub Actions?
Because I tried to push a new test repo(a2p_ghp_test) under your GH, to test the default branch `gh-pages`. But now the log shows an error that the env does not have the GitHub token to push to a new repo- 
```
══ Failed tests ════════════════════════════════════════════════════════════════
── Error ('test-compiler-ghpages.R:57:3'): animint2pages() default branch is gh-pages ──
Error in `value[[3L]](cond)`: A GitHub token is required to create and push to a new repository. 
To create a GitHub token, follow these steps:
1. Go to https://github.com/settings/tokens/new?scopes=repo&description=animint2pages
2. Confirm your password if prompted.
3. Ensure that the 'repo' scope is checked.
4. Click 'Generate token' at the bottom of the page.
5. Copy the generated token.
After creating the token, you can set it up in your R environment by running: 
Sys.setenv(GITHUB_PAT="yourGithubPAT") 
gert::git_config_global_set("user.name", "yourUserName") 
gert::git_config_global_set("user.email", "yourEmail") 

Backtrace:
    ▆
 1. └─animint2::animint2pages(viz, github_repo = "a2p_ghp_test") at test-compiler-ghpages.R:[57](https://github.com/animint/animint2/actions/runs/9869578505/job/27253578004#step:11:58):3
 2.   └─base::tryCatch(...)
 3.     └─base (local) tryCatchList(expr, classes, parentenv, handlers)
 4.       └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
 5.         └─value[[3L]](cond)
```

based on this comment https://github.com/animint/animint2/pull/101#issuecomment-1793237082, the animint2pages_test_repo seems has been created before generating the token?

tdhock:
yes that is right, I created the repo using github web interface, then followed those instructions to create PAT which gives animint/animint2 repo permissions to push to tdhock/animint2pages_test_repo during gh actions testing of animint2pages.

tdhock:
Why can't we just keep using animint2pages_test_repo, with the same PAT, and just add a new test to make sure that the default branch that we get when we clone it is gh-pages?

tests/testthat/test-compiler-ghpages.R
```diff
@@ -45,3 +45,18 @@ test_that("animint2pages raises an error if no GitHub token is present", {
...
+    owner <- "tdhock"
+  } else {
+    owner <- whoami[["login"]]
+  }
+  local_repo_path <- tempfile(pattern = "repo_clone_")
+  test_viz_repo <- animint2pages(viz, github_repo = "a2p_ghp_test")
+  gert::git_clone(url = paste0("https://github.com/", owner, "/a2p_ghp_test.git"), path = local_repo_path)
+  # Check the default branch after clone
+  default_branch <- gert::git_branch(repo = local_repo_path)
+  expect_equal(default_branch, "gh-pages")
```
tdhock, line 61: can this test be added inside the "animint2pages() returns owner/repo string" block above?

Faye-yufan:
> Why can't we just keep using animint2pages_test_repo, with the same PAT, and just add a new test to make sure that the default branch that we get when we clone it is gh-pages?

Yes we can just use animint2pages_test_repo only. The reason I added a new repo here is just for testing on GH Actions, because animint2pages() function itself cannot change the default branch of an existing GH repo. The test worked on my local, but I wanted to create a new repo with updated animint2pages(), which sets`gh-pages` as the default branch, to see if the test works on GH Actions.

Faye-yufan:
@tdhock Could you create a new animint2pages_test_repo using this branch version of `animint2pages`? Or maybe I can try create a test repo under animint org to see if it works.

tdhock:
"Could you create a new animint2pages_test_repo using this branch version of animint2pages?" 
I could but I'm not sure that would address this testing issue.
the issue is that the current PAT only has write access to one repo, which is already created, but we want to test creation of a new repo, right?
can we create a new org,  "animint-testing" and make a PAT that has permission to create new repos in that org?

tdhock:
> can we create a new org, "animint-testing" and make a PAT that has permission to create new repos in that org?

this was implemented in #147 which also required permission to delete repos.

PR #144, opened 2024-08-27
rm selenium, combine renderer1,2,3

tdhock:
hi @siddhesh195 
this is a follow-up to #126 

* removed RSelenium in testthat.R and DESCRIPTION
* run only one renderer CI job instead of three


DESCRIPTION
```diff
@@ -87,7 +87,7 @@ Suggests:
     htmltools,
     rmarkdown,
     testthat,
-    RSelenium, XML,
```
siddhesh195, line 90: We also do not need to start selenium docker in tests.yaml. I think the build is currently failing on Github because of that

siddhesh195 (approved):

DESCRIPTION
tdhock: ok fixed

tdhock:
looks like build is working now with renderer tests all combined into a single CI job, and with Selenium removed.
Do you see any other places in the code where we could remove selenium?

PR #182, opened 2025-01-24
fix guided tour for variable/value aes (named clickSelects/showSelected)

tdhock:
The PR improves the DriverJS functionality.
Before this PR the DriverJS description did not include selector names, for named clickSelects/showSelected.
This PR changes the compiler to add help_clickSelects and help_showSelected parameters to each geom in the plot.json, which include the selector names, taken from the data values in the case of named clickSelects/showSelected.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: animint2
 Title: Animated Interactive Grammar of Graphics
-Version: 2025.1.24
+Version: 2025.1.25
 URL: https://animint.github.io/animint2/
 BugReports: https://github.com/animint/animint2/issues
 Authors@R: c(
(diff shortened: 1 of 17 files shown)


tdhock:
this PR also adds an argument to animint2pages, chromote_sleep_seconds=NULL (new default) means to not try to do auto screen capture, which seems to not be working...
```r
> animint2pages(wb.facets, "2025-01-WorldBank-facets-map")
Le chargement a nécessité le package : gert
Le chargement a nécessité le package : gh
Le chargement a nécessité le package : chromote
Le chargement a nécessité le package : magick
Le chargement a nécessité le package : grDevices
Serving the directory /tmp/RtmpXQyylH/file92db27ba9c98 at http://127.0.0.1:4025
Erreur : R: geometry does not contain image `/tmp/RtmpXQyylH/file92db27ba9c98/Capture_full.PNG' @ warning/attribute.c/GetImageBoundingBox/519
```

PR #184, opened 2025-01-26
hello world example in ?animint

tdhock:
TODO increase size of "a" inside legend svgs, right now text size is 3.88, why so small?

image


tdhock:
probably need to fix this line in getLegendList:
```r
    gdefs[[leg]]$text_size <- getTextSize("legend.text", theme)
```

Issue #179, opened 2025-01-19
geom_text(aes(tooltip)) does not work


tdhock:
because tooltips are currently implemented by adding <title> elements inside the svg element, but the JS `.text` method is removing the <title> element and replacing with the aes(label) value.
we could conceivably get it to work by making html like <text>label text<title>tooltip text</title></text>
but even better is if we could implement a new tooltip library with more rapid/obvious tooltips that show up.

suhaani-agarwal:
[Tippy.js](https://atomiks.github.io/tippyjs/) can be used for this......can I work on it?

tdhock:
great idea, please go ahead

tdhock:
in your pr make sure to remove the existing tooltip implementation via <title> elements, so the new implementation works with all geoms.

PR #191, opened 2025-03-20
Added tooltip in aes using tippy.js

suhaani-agarwal:
Resolves issue #179 

I have added the tooltip parameter for aes and this is the initial PR for the same. 
This is the video demonstration of the output that I got on rendering a viz using this tooltip implementation:
https://youtu.be/gAa8a4jaKMk

- Right now the index.html contains definite links but I plan to make them relative like the others inside vendor/ 

- If you approve this, I will write a test for this tooltip implementation.

PR #197, opened 2025-06-01
Adding aes(tooltip) using d3

suhaani-agarwal:
Resolves issue #179 

 - This PR changes the tooltip system from tippy.js (being implemented in PR #191 ) to a simpler D3 version. The tippy.js implementation was causing two main problems. First, it sometimes made the browser slow or unresponsive and caused chromote timeout issues , especially with many tooltips (hence certain tests were failing) . Second, it required adding extra JavaScript files that made the package bigger.

 - Since animint already uses D3 for everything else, I think it makes sense to use it for tooltips too. Here, This implementation creates a single tooltip div that gets added to the plot container. When hovering over any element with aes(tooltip) or clickSelects / showSelected , the tooltip's  position updates to follow the mouse and its content changes to show the specific tooltip text from the hovered element

 - The same div is reused for all tooltips - only its content and position change based on which element is being hovered. This avoids creating multiple tooltip elements, keeps everything in one place and also makes the tooltips faster . The tooltip hides automatically when moving the mouse away.
 
 - I have also modified the existing tooltip tests to support the new tooltip implementation.
 
 Is this implementation approach okay?

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1610,6 +1610,20 @@ var animint = function (to_select, json_file) {
     };
     doActions(enter);  // DO NOT DELETE!
     var has_tooltip = g_info.aes.hasOwnProperty("tooltip");
+    function positionTooltip(tooltip, content) {
+      var mouseX = 0, mouseY = 0;
+      try {
+        mouseX = (d3.event && d3.event.pageX) || 0;
+        mouseY = (d3.event && d3.event.pageY) || 0;
+      } catch (e) {
+        console.error("Error getting mouse position:", e);
+      }
+      tooltip
+        .html(content)
+        .style("left", (mouseX + TOOLTIP_HORIZONTAL_OFFSET) + "px")
+        .style("top", (mouseY - TOOLTIP_VERTICAL_OFFSET) + "px")
+        .style("opacity", 1);
+    }
     if(has_clickSelects || has_tooltip || has_clickSelects_variable){
       // Tooltip positioning constants
       var TOOLTIP_HORIZONTAL_OFFSET = 10; // pixels right of mouse pointer
@@ -1646,34 +1660,14 @@ var animint = function (to_select, json_file) {
             console.error("Error generating tooltip content:", e);
             return;
           }
-          var mouseX = 0, mouseY = 0;
-          try {
-            mouseX = (d3.event && d3.event.pageX) || 0;
-            mouseY = (d3.event && d3.event.pageY) || 0;
-          } catch(e) {
-            console.error("Error getting mouse position:", e);
-          }
-          tooltip
-            .html(content)
-            .style("left", (mouseX + TOOLTIP_HORIZONTAL_OFFSET) + "px")
-            .style("top", (mouseY - TOOLTIP_VERTICAL_OFFSET) + "px")
-            .style("opacity", 1);
+          positionTooltip(tooltip, content);
         })
         .on("mouseout.tooltip", function() {
           tooltip.style("opacity", 0)
           .html(null);
         })
         .on("mousemove.tooltip", function() {
-          var mouseX = 0, mouseY = 0;
-          try {
-            mouseX = (d3.event && d3.event.pageX) || 0;
-            mouseY = (d3.event && d3.event.pageY) || 0;
-          } catch(e) {
-            console.error("Error getting mouse position:", e);
-          }
-          tooltip
-            .style("left", (mouseX + TOOLTIP_HORIZONTAL_OFFSET) + "px")
-            .style("top", (mouseY - TOOLTIP_VERTICAL_OFFSET) + "px");
+          positionTooltip(tooltip, tooltip.html());
         });
     }
     if(Selectors.hasOwnProperty(selector_name)){
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1656,6 +1656,8 @@ var animint = function (to_select, json_file) {
         })
         .on("mouseout.tooltip", function() {
           tooltip.style("opacity", 0)
+          .style("left", null)
+          .style("top", null)
           .html(null);
         })
         .on("mousemove.tooltip", function() {
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1611,6 +1611,10 @@ var animint = function (to_select, json_file) {
     doActions(enter);  // DO NOT DELETE!
     var has_tooltip = g_info.aes.hasOwnProperty("tooltip");
     if(has_clickSelects || has_tooltip || has_clickSelects_variable){
+      // Tooltip positioning constants
+      var TOOLTIP_HORIZONTAL_OFFSET = 10; // pixels right of mouse pointer
+      var TOOLTIP_VERTICAL_OFFSET = 28;   // pixels above mouse pointer
+
       var text_fun;
       if(has_tooltip){
         text_fun = function(d){
@@ -1653,8 +1657,8 @@ var animint = function (to_select, json_file) {
           }
           tooltip
             .html(content)
-            .style("left", (mouseX + 10) + "px")
-            .style("top", (mouseY - 28) + "px")
+            .style("left", (mouseX + TOOLTIP_HORIZONTAL_OFFSET) + "px")
+            .style("top", (mouseY - TOOLTIP_VERTICAL_OFFSET) + "px")
             .style("opacity", 1);
         })
         .on("mouseout.tooltip", function() {
@@ -1670,8 +1674,8 @@ var animint = function (to_select, json_file) {
             console.error("Error getting mouse position:", e);
           }
           tooltip
-            .style("left", (mouseX + 10) + "px")
-            .style("top", (mouseY - 28) + "px");
+            .style("left", (mouseX + TOOLTIP_HORIZONTAL_OFFSET) + "px")
+            .style("top", (mouseY - TOOLTIP_VERTICAL_OFFSET) + "px");
         });
     }
     if(Selectors.hasOwnProperty(selector_name)){
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1612,11 +1612,9 @@ var animint = function (to_select, json_file) {
     var has_tooltip = g_info.aes.hasOwnProperty("tooltip");
     function positionTooltip(tooltip, content) {
       var mouseX = 0, mouseY = 0;
-      try {
-        mouseX = (d3.event && d3.event.pageX) || 0;
-        mouseY = (d3.event && d3.event.pageY) || 0;
-      } catch (e) {
-        console.error("Error getting mouse position:", e);
+      if (d3.event) {
+        mouseX = d3.event.pageX;
+        mouseY = d3.event.pageY;
       }
       tooltip
         .html(content)
@@ -1653,13 +1651,7 @@ var animint = function (to_select, json_file) {
       elements
         .on("mouseover.tooltip", function(d) {
           if (!d || typeof text_fun !== 'function') return;
-          var content;
-          try {
-            content = text_fun(d);
-          } catch(e) {
-            console.error("Error generating tooltip content:", e);
-            return;
-          }
+          var content = text_fun(d);
           positionTooltip(tooltip, content);
         })
         .on("mouseout.tooltip", function() {
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1635,8 +1635,6 @@ var animint = function (to_select, json_file) {
     : d3.select("#plot").append("div")
         .attr("class", "animint-tooltip")
         .style("opacity", 0);
-      // Remove existing title elements to avoid conflicts with tooltips
-      elements.selectAll("title").remove();
       // Add tooltip handlers
       elements
         .on("mouseover.tooltip", function(d) {
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1626,11 +1626,53 @@ var animint = function (to_select, json_file) {
 	  return d["clickSelects.variable"] + " " + d["clickSelects.value"];
 	};
       }
-      // if elements have an existing title, remove it.
+      var tooltip = d3.select("#plot").select(".animint-tooltip").node() 
+    ? d3.select(".animint-tooltip")
+    : d3.select("#plot").append("div")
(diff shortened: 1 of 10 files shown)


tests/testthat/test-renderer1-tooltip.R
```diff
@@ -56,51 +56,63 @@ subset(years, year==1975)
...
-
-test_that("line tooltip renders as title", {
-  info <- animint2HTML(viz)
-  title.nodes <- getNodeSet(info$html, '//g[@class="geom1_line_linetip"]//title')
-  value.vec <- sapply(title.nodes, xmlValue)
-  expect_identical(value.vec, c("group 1", "group 2"))
+test_that("tooltip shows correct content for point", {
+  remDr$Input$dispatchMouseEvent(
+    type = "mouseMoved", 
+    x = 230.01787959856264, #coordinates for the circle corresponding to Country Myanmar
```
tdhock, line 101: I don't think it is a good idea to use these values in the test.  could you please search for id="Myanmar" instead?

tests/testthat/test-renderer1-tooltip.R
```diff
@@ -82,8 +82,9 @@ test_that("tooltip shows correct content for rect", {
   center_y <- rect_y + (rect_height / 2)
   
   # Simulate hover over rectangle center
-  remDr$Input$dispatchMouseEvent(type = "mouseMoved", x = center_x, y = center_y)
-  Sys.sleep(0.5) # Wait for tooltip
+  remDr$Input$dispatchMouseEvent(type = "mouseMoved", x = center_x + 10, y = center_y)
```
tdhock, line 85: please remove the +10 here, which is another magic number. (and below, and +5 etc)
instead can you please try using something like https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect ?
```r
runtime_evaluate(script=...getBoundingClientRect())
```
etc

inst/htmljs/animint.js
```diff
@@ -1626,11 +1626,53 @@ var animint = function (to_select, json_file) {
...
+        .on("mousemove.tooltip", function() {
+          var mouseX = 0, mouseY = 0;
+          try {
+            mouseX = (d3.event && d3.event.pageX) || 0;
+            mouseY = (d3.event && d3.event.pageY) || 0;
+          } catch(e) {
+            console.error("Error getting mouse position:", e);
+          }
+          tooltip
+            .style("left", (mouseX + 10) + "px")
```
tdhock, line 1673: the 10 and 28 here are examples of "magic numbers" -- can you please put them in a variable that would clarify their interpretation?
This is not exactly a violation of "repeating" but it is similar to this entry in my general usage rubric:
-5 for repeating constants in your code. (this is error-prone, and confusing for readers of your code) solution: declare a variable with that constant and use that variable several times. (the name of the constant should document its purpose/usage) Think: if I wanted to change that constant, in how many places would I need to change it? (the answer should be one) See discussion https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

suhaani-agarwal: > the 10 and 28 here are examples of "magic numbers" -- can you please put them in a variable that would clarify their interpretation? This is not exactly a violation of "repeating" but it is similar to this entry in my general usage rubric: -5 for repeating constants in your code. (this is error-prone, and confusing for readers of your code) solution: declare a variable with that constant and use that variable several times. (the name of the constant should document its purpose/usage) Think: if I wanted to change that constant, in how many places would I need to change it? (the answer should be one) 

Thank you for pointing that out , I see what you mean about the magic numbers and how naming those constants makes the code easier to understand and maintain.  I’ve replaced the numbers with descriptive variables now, I'll definitely keep this practice in mind moving forward.

inst/htmljs/animint.js
```diff
@@ -1611,6 +1611,10 @@ var animint = function (to_select, json_file) {
     doActions(enter);  // DO NOT DELETE!
     var has_tooltip = g_info.aes.hasOwnProperty("tooltip");
     if(has_clickSelects || has_tooltip || has_clickSelects_variable){
+      // Tooltip positioning constants
+      var TOOLTIP_HORIZONTAL_OFFSET = 10; // pixels right of mouse pointer
+      var TOOLTIP_VERTICAL_OFFSET = 28;   // pixels above mouse pointer
```
tdhock, line 1616: yes!!

tests/testthat/test-renderer3-prostateLasso.R
```diff
@@ -100,16 +100,7 @@ html.click2 <- getHTML()
...
-  "(() => {
-      const rect = document.querySelector('g.geom3_tallrect_path rect#arclength101');
-      const box = rect.getBoundingClientRect();
-      return {
-        left: box.left,
-        top: box.top
-      };
-  })()"
-  )
+  tallrect_position <- get_element_bbox("g.geom3_tallrect_path rect#arclength101")
```
tdhock, line 103: yes!!

tdhock:
overall this is really excellent work, thank you!
can you please add a test case for geom_text?

tests/testthat/test-renderer1-tooltip.R
```diff
@@ -9,12 +9,12 @@ viz <-
...
        geom_point(aes(life.expectancy, fertility.rate,
                       colour=region, size=population,
-                      tooltip=paste(country, "population", population),
+                      tooltip=paste(country, "population", population), id = country,
                       key=country), # key aesthetic for animated transitions!
                   clickSelects="country",
                   showSelected="year",
                   data=WorldBank)+
-       geom_text(aes(life.expectancy, fertility.rate, label=country,
+       geom_text(aes(life.expectancy, fertility.rate, label=country, tooltip = country, id = country,
```
tdhock, line 17: this will produce invalid HTML.
id must be unique across all elements on the page.
please change, for example `aes(id=paste0("text_",country))`

inst/htmljs/animint.js
```diff
@@ -1626,11 +1630,51 @@ var animint = function (to_select, json_file) {
...
+            mouseX = (d3.event && d3.event.pageX) || 0;
+            mouseY = (d3.event && d3.event.pageY) || 0;
+          } catch(e) {
+            console.error("Error getting mouse position:", e);
+          }
+          tooltip
+            .html(content)
+            .style("left", (mouseX + TOOLTIP_HORIZONTAL_OFFSET) + "px")
+            .style("top", (mouseY - TOOLTIP_VERTICAL_OFFSET) + "px")
+            .style("opacity", 1);
```
tdhock, line 1660: this block is almost repeated below.
can you please make a function that is used twice, in order to emphasize the similarity?

inst/htmljs/animint.js
```diff
@@ -1610,6 +1610,20 @@ var animint = function (to_select, json_file) {
     };
     doActions(enter);  // DO NOT DELETE!
     var has_tooltip = g_info.aes.hasOwnProperty("tooltip");
+    function positionTooltip(tooltip, content) {
+      var mouseX = 0, mouseY = 0;
+      try {
+        mouseX = (d3.event && d3.event.pageX) || 0;
+        mouseY = (d3.event && d3.event.pageY) || 0;
+      } catch (e) {
+        console.error("Error getting mouse position:", e);
```
tdhock, line 1619: can this error be removed?

inst/htmljs/animint.js
```diff
@@ -1610,6 +1610,20 @@ var animint = function (to_select, json_file) {
     };
     doActions(enter);  // DO NOT DELETE!
     var has_tooltip = g_info.aes.hasOwnProperty("tooltip");
+    function positionTooltip(tooltip, content) {
+      var mouseX = 0, mouseY = 0;
+      try {
+        mouseX = (d3.event && d3.event.pageX) || 0;
```
tdhock, line 1616: is this `|| 0` redundant with the `mouseX = 0` above? can it be removed?

inst/htmljs/animint.js
```diff
@@ -1610,6 +1610,20 @@ var animint = function (to_select, json_file) {
...
+        mouseY = (d3.event && d3.event.pageY) || 0;
+      } catch (e) {
+        console.error("Error getting mouse position:", e);
+      }
+      tooltip
+        .html(content)
+        .style("left", (mouseX + TOOLTIP_HORIZONTAL_OFFSET) + "px")
+        .style("top", (mouseY - TOOLTIP_VERTICAL_OFFSET) + "px")
+        .style("opacity", 1);
```
tdhock, line 1625: I see this opacity 1 was only present in one of the two versions below, is that important to keep? or is it ok to do opacity 1 in both now?

inst/htmljs/animint.js
suhaani-agarwal: > is this `|| 0` redundant with the `mouseX = 0` above? can it be removed?

yes, I have removed it now

tests/testthat/test-renderer3-tooltip-interactivity.R
```diff
@@ -4,56 +4,38 @@ data("CO2")
...
-  
-  title_nodes2 <-
-    getNodeSet(info$html_updated1, '//g[@class="geom1_point_p"]//circle//title')
-  rendered_titles2 <- sapply(title_nodes2, xmlValue)
-  rendered_titles2_unique <- unique(rendered_titles2)
-  actual_titles2 <- c("Qc1", "Qc2", "Qc3", "Mc1", "Mc2", "Mc3")
-  
-  expect_equal(length(rendered_titles2), sum(CO2$Treatment == "chilled"))
-  expect_identical(sort(rendered_titles2_unique), sort(actual_titles2))
-
```
tdhock, line 45: are we losing any test coverage by deleting this treatment chilled interaction test?
if not, can you please justify deleting this test?

tdhock: (it is ok as long as the functionality is tested somewhere else, for example one of your new test cases. but I just do not understand what this test is doing and why we may want to keep it)

tdhock: I like deleting the variable names with numbers in them, `title_nodes3` etc

suhaani-agarwal: I've added a new test (Interactivity does not mess up tooltips) that verifies tooltip correctness for both chilled and nonchilled points after toggling visibility via clicks (something like the old test) . It uses uniquely generated ids per circle (paste0(Plant, "_", seq_len(nrow(CO2))))

Also, the tooltip hover behavior is already tested in renderer1-tooltip.R, so the second test in this file ("tooltip shows correct content on hover interaction") can be removed. Let me know if I should remove it.

Does the new test look good, or would you suggest any changes?

tests/testthat/test-renderer3-tooltip-interactivity.R
```diff
@@ -3,7 +3,7 @@ acontext("tooltip-interactivity")
 data("CO2")
 
 plot_viz <- ggplot() + 
-  geom_point(aes(conc, uptake, color=Treatment, tooltip=Plant),
+  geom_point(aes(conc, uptake, color=Treatment, tooltip=Plant , id = paste0(CO2$Plant, "_", seq_len(nrow(CO2)))), # Unique ID for every circle
```
tdhock, line 6: please remove `CO2$`
in general for `geom_(data=df)` please avoid `aes(visual_property=df$variable)` because you can write `aes(visual_property=variable)` instead (simpler)

suhaani-agarwal: done!

tests/testthat/test-renderer1-tooltip.R
```diff
@@ -56,51 +56,67 @@ subset(years, year==1975)
...
-  nodes <-
-    getNodeSet(info$html, '//g[@class="geom1_point_scatter"]//circle//title')
-  tooltips <- sapply(nodes, xmlValue)
-  expect_match(tooltips, "population")
+test_that("animint-tooltip div exists with correct initial state", {
+  tooltip_div <- getNodeSet(info$html, '//div[@class="animint-tooltip"]')
+  expect_equal(length(tooltip_div), 1)
+  # Check initial opacity is 0
+  style <- xmlGetAttr(tooltip_div[[1]], "style")
+  expect_match(style, "opacity: 0;")
```
tdhock, line 64: can you please change this to expect_style()?
```
grep --color=auto -nH --null -e expect_style *R
helper-functions.R:154:expect_styles <- function(html, styles.expected){
test-renderer1-rect-size.R:16:  expect_styles(info$html, list("stroke-width"="^5[a-z]*$"))
test-renderer1-rect-size.R:26:  expect_styles(info$html, list("stroke-width"="^0.01[a-z]*$")) 
test-renderer1-rect-size.R:35:  expect_styles(info$html, list("stroke-width"=c("^0.01[a-z]*$", "^5[a-z]*$")))
test-renderer1-segment-size.R:15:  expect_styles(info$html, list("stroke-width"="^1[a-z]*$"))
test-renderer1-segment-size.R:24:  expect_styles(info$html, list("stroke-width"=c("^5[a-z]*$", "^10[a-z]*$")))
```

tests/testthat/test-renderer1-tooltip.R
```diff
@@ -111,15 +127,9 @@ ex_plot <- ggplot() +
...
-  # See that every <a> element has a title (the country name) initially
-  title_nodes <-
-    getNodeSet(info$html, '//g[@class="geom1_point_ex"]//a//title')
-  rendered_titles <- sapply(title_nodes, xmlValue)
-  expect_identical(sort(rendered_titles), sort(expected.countries))
+test_that("tooltip div exists with href elements", {
+  tooltip_div <- getNodeSet(info$html, tooltip.xpath)
+  expect_equal(length(tooltip_div), 1)
+  opacity <- getStyleValue(info$html, tooltip.xpath, "opacity")
+  expect_identical(opacity, "0")
```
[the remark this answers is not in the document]
suhaani-agarwal: It is correct that the opacity is 1 when the tooltip is shown and 0 when tooltip is not shown. 
Opacity 1 is being checked in test-renderer3-tooltip-interactivity after mouseHover

I have now added opacity expectations in [test-renderer1-tooltip.R](https://github.com/animint/animint2/pull/197/files#diff-40efed45ce99199e4cd7f85d190cf3cb29981e9b05f939b89a13f53701238ea4) too as follows
 - expect_identical(opacity, "1") right after hover,
 - and expect_identical(opacity, "0") after mouse out. (for all 3 hover cases being implemented in the test)

Is that okay?

PR #153, opened 2024-10-08
grid plot layout via theme_animint(rowspan, colspan, last_in_row)

tdhock:
closes #115 

this proposal supersedes #139 which should be closed if this PR is sufficient.

The idea is to define three new plot attributes, to be specified inside `theme_animint`

* `rowspan=2` means plot takes up two rows (default=1)
* `colspan=2` means plot takes up two columns (default=1)
* `last_in_row=TRUE` means this plot ends a row, HTML `<tr>` element. (default=FALSE)

the R code compiler needs to save these as attributes of the corresponding plot list.

For example
```r
animint(
  ggplot()+theme_animint(rowspan=2)+ggtitle("first td in first tr, rowspan=2"),
  ggplot()+theme_animint(last_in_row=TRUE)+ggtitle("second td in first tr"),
  ggplot()+ggtitle("first td in second tr")
)
```
should make HTML like below
image

if any of these three attributes are specified in any of the ggplots, then the renderer JavaScript code puts the plot in the corresponding tr and td of the "outer table," according to the following rules:

* we create a new `<tr>` before the first plot, and after every plot with `last_in_row=TRUE`
* for every plot, we create a new `<td>` inside the current `<tr>` 

There is no need for computing the total/max number of rows/columns.

@siddhesh195 please consider this PR and if you agree, close the other PR, and move any relevant code to this branch (compiler R code, tests, etc).

PR #199, opened 2025-06-03
Feature rowspan, colspan

### LINKED ISSUE #115: layout ggplots in a table
currently ggplots appear one after another in the source code, and the visual layout is determined interactively, by resizing/zooming the window. layout here means how to arrange the ggplots in a grid (is plot 3 to the right of plot 2, or below plot 1?)
would be nice if we could implement some simple method for defining the layout.
for example below plots are in a 2x2 table, plot1 is on the first/left column, spanning both rows; plot2 is on the second/right column, first row; plot3 is on the third/right column, second row.
```r
animint(
  plot1=ggplot()+theme_animint(row=1, col=1, rowspan=2)),
  plot2=ggplot()+theme_animint(row=1, col=2),
  plot3=ggplot()+theme_animint(row=2, col=2))
```

biplab-sutradhar:
Closes #115 

This pr is about the rowspan and colspan features and will be updated with tests and code refactoring.

### DIFF
--- a/R/z_animint.R
+++ b/R/z_animint.R
@@ -200,6 +200,7 @@ parsePlot <- function(meta, plot, plot.name){
   options_list <- getWidthAndHeight(plot$theme)
   options_list <- setUpdateAxes(plot$theme, options_list)
   plot.info$options <- options_list
+  plot.info$attributes <- theme_attribute(plot$theme)
 
   list(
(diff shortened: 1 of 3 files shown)


tdhock:
instead of creating this new PR, why not modify #153 ?

PR #201, opened 2025-06-06
animint2pages() create dir, remove err file, tests

### LINKED ISSUE #134: set default branch of new repo in animint2pages

### LINKED ISSUE #202: animint2pages removes Capture.png
sometimes we create an animint2pages()
and then we add it to a gallery, which entails adding a Capture.PNG file to the gh-pages branch.
The next time we run animint2pages(), the Capture.PNG file is removed.
The next time we add that viz to a gallery, we need to re-add the Capture.PNG file, which is inconvenient.
It would be more user-friendly if animint2pages checks for existence of Capture.PNG and keeps the old one, instead of removing it.

tdhock:
Closes #202 
Closes #134


tdhock:
@suhaani-agarwal 
@biplab-sutradhar 
can you please review this PR?

Issue #174, opened 2024-12-03
direct labels aligned boxes qp solver


tdhock:
in R we have library(quadprog) which we use in directlabels::last.polygons etc https://tdhock.github.io/directlabels/docs/lineplot/posfuns/last.polygons.html
to do that in javascript we could use https://github.com/albertosantini/quadprog
it would probably be better to introduce a new geom for that, geom_label_column etc? with aes x, y, color, fill, label (top/bottom of box calculated based on label text and y)

PR #203, opened 2025-06-06
implementing geom_label_aligned

suhaani-agarwal:
resolving #174 

This PR introduces a new geom, geom_aligned_boxes, as proposed in the issue. (name of the geom can be changed if needed)

Status: Initial implementation — still a work in progress and requires a lot of modifications and bug fixing. I am pushing the code made till now to seek guidance.
Current issue: While testing with examples, I noticed that the generated HTML renders:

``<g class="geom1_alignedboxes_plot"><g class="PANEL1"></g></g>``

The PANEL1 elements are empty, no boxes are being rendered. I suspect I may be missing a key step in correctly registering or drawing the new geom. Any guidance on what might be causing this?


R/geom-aligned-boxes.R
```diff
@@ -0,0 +1,174 @@
...
+  result <- quadprog::solve.QP(Dmat, dvec, t(Amat), bvec)
+  return(result$solution)
+}
+
+# Helper function to optimize horizontal positions using quadratic programming
+optimize_horizontal_positions <- function(x, widths, min_distance) {
+  n <- length(x)
+  if (n <= 1) return(x)
+  Dmat <- diag(n)
+  dvec <- x
```
tdhock, line 160: these need to move to javascript I think

tdhock: because we only know how big the text is (in pixels) after drawing it on screen.
and we need to know text size to avoid overlaps.

suhaani-agarwal:
I have implemented the new geom -  geom_aligned_boxes, which takes x, y, and label as required aesthetics, along with optional ones like fill, stroke, and color etc. On the JavaScript side, it dynamically measures the text size and draws a rectangle around it positioned by quadprog on renderer side, to form a label box. an example of the structure generated in the rendered HTML is as follows:
```
<g class="geom1_alignedboxes_pricesPlot">
  <g class="PANEL1">
    <g class="geom" style="opacity: 1; stroke: black; stroke-width: 2; fill: rgb(248, 118, 109);">
      <rect x="135.0775" y="226.1861" width="50.9219" height="28.5156" stroke-width="1" rx="0" ry="0"></rect>
      <text x="160.5385" y="244.4439" font-size="12px" style="text-anchor: middle;" stroke-width="1">Apple</text>
    </g>
  </g>
</g>
```

Although most of the initial implementation is complete, the positioning of these boxes using quadprog.js is not yet accurate. The elements are not restricted to the axis boundaries. I plan to refine the constraints in the quadratic programming setup to ensure correct and bounded placement. Once that is done, I will make the renderer tests for this geom.

Please let me know if the current structure and approach are appropriate or if any modifications are required.

inst/htmljs/vendor/quadprog.js
```diff
@@ -0,0 +1,600 @@
+/**
+ * JavaScript implementation of quadratic programming solver
+ * Based on the algorithm from quadprog package
+ */
```
tdhock, line 4: where did this code come from?
please add a link / explanation

tdhock: I understand that you copied the code from https://github.com/albertosantini/quadprog/tree/master/lib -- please add a link/reference.
"package quadprog" means a JS package, not R package, right? please make sure that is clear?

suhaani-agarwal:
I am still working on some edge cases for this positioning function that uses quadratic programming. (like examples with horizontal alignment are not yet being positoned in the most optimal way)

inst/htmljs/vendor/quadprog.js
```diff
@@ -598,75 +601,213 @@ function solveQP(Dmat, dvec, Amat, bvec = [], meq = 0, factorized = [0, 0]) {
...
-    console.log("got data : ", data, "alignement : ", alignment)
-  var n = data.length;
+// ---------------------------------------------------------------------------------------------------
+// Label positioning functions inspired by the R `directlabels` package: https://github.com/tdhock/directlabels
+//
+// `optimizeAlignedBoxes` uses a QP solver to reposition only the overlapping label boxes,
+// finding the nearest possible positions to their original locations, subject to the constraint
+// that the boxes do not overlap.
+//
+// `bumpup` and `bumpleft` are heuristic fallback strategies, also inspired by directlabels.
```
tdhock, line 611: i would prefer to not add bumpup and bumpleft. can you please remove? or justify their addition?

suhaani-agarwal:
An example of how geom_aligned_boxes look is as follows : 

https://github.com/user-attachments/assets/73a00414-eb31-420e-a98d-d58aeb12836c

 - I have to work on how the text looks like. right now it looks a little extra bold. I will correct that.
 - I may have to work on more robust boundary constraints
 - I will add this example as a test (testing that the geom aligned boxes do not collide even after interactions and some other aspects of this geom)

inst/examples/WorldBank-facets-map.R
```diff
@@ -90,7 +90,7 @@ wb.facets <- animint(
       size=4,
       alpha=1,
       alpha_off=0.1)+
-    geom_text(aes(
+    geom_aligned_boxes(aes(
```
tdhock, line 93: I tried geom_aligned_boxes in the world bank data viz and I am very impressed at how it "just works" !!
image
However there are a few issues.

* I was expecting white background by default (not black).
* this geom specifies hjust=1 (right justify) but we see hjust=0.5 (center justify)
* the horizontal size seems to be incorrect (Trinidad and Tobago has a lot of extra space in the box)

This is a really excellent first try though!

Can we change the name from geom_aligned_boxes to geom_label_something? for consistency with geom_label in ggplot2? maybe geom_label_aligned or geom_label_no_overlap?

tests/testthat/test-renderer1-geom-aligned-boxes.R
```diff
@@ -0,0 +1,181 @@
...
+      width = as.numeric(attrs["width"]),
+      height = as.numeric(attrs["height"])
+    )
+  })
+  for (i in 1:(length(box_info)-1)) {
+    for (j in (i+1):length(box_info)) {
+      box1 <- box_info[[i]]
+      box2 <- box_info[[j]]
+      x_overlap <- box1$x < (box2$x + box2$width) && (box1$x + box1$width) > box2$x
+      y_overlap <- box1$y < (box2$y + box2$height) && (box1$y + box1$height) > box2$y
```
tdhock, line 94: this logic is repeated below. please write a function and use it in both cases to avoid repetitive code.

inst/examples/WorldBank-facets-map.R
suhaani-agarwal: > * I was expecting white background by default (not black).

I have now changed the default fill to white.

> Can we change the name from geom_aligned_boxes to geom_label_something? for consistency with geom_label in ggplot2? maybe geom_label_aligned or geom_label_no_overlap?

I have now renamed the geom to geom_label_aligned

R/geom-label-aligned.R
```diff
@@ -0,0 +1,147 @@
+#' Non-overlapping label boxes
+#'
+#' This geom creates boxes with labels that are aligned either vertically or horizontally,
+#' using quadratic programming to optimize their positions and avoid overlaps.
```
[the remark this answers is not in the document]
suhaani-agarwal: - QP is used after all data filtering and grouping, including showSelected, has been applied.
 - QP is used to optimize the positions of label boxes to avoid overlaps.
The optimization is performed along one axis only:
If alignment = "vertical", QP is used to adjust the Y positions of the labels (so they don't overlap vertically).
If alignment = "horizontal", QP is used to adjust the X positions of the labels (so they don't overlap horizontally).
 - The other axis (X for vertical, Y for horizontal) is kept fixed.
 - The alignment param can be specified by the user as "horizontal" or "vertical" while writing the geom. 
By default, the alignment is kept "vertical". 

 - I have now added information about the QP in geom-label-aligned.R in the comments.

inst/examples/geom_aligned_boxes_example.R
```diff
@@ -0,0 +1,88 @@
...
+
+viz3 <- list(
+  overlapTest = ggplot() +
+    geom_aligned_boxes(aes(x, y, label = label, fill = group), data = label_data) +
+    ggtitle("Overlap Test: Aligned Boxes") +
+    xlab("X") +
+    ylab("Y")
+)
+
+animint2dir(viz3, "overlap-aligned-boxes")
```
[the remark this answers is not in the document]
suhaani-agarwal: This particular data viz is not an ideal example and I have now replaced it with the interactive data viz using world bank data that I was using for the tests. 
 - The user can specify on which axis QP will be run by the "alignment" parameter as specified here : [geom-label-aligned.R](https://github.com/animint/animint2/pull/203/files#diff-77b2d748ddd9a67b76a65f60bf619769fbccb63dbadd24cc53945bade93c85b1:~:text=%40%40%20%2D0%2C0%20%2B1%2C161%20%40%40-,%23%27%20Non%2Doverlapping%20label%20boxes,%23%27,-%23%27%20%40inheritParams%20layer) in the comments. 
 
> Is the QP run for all data points? I think there are some use cases where it would be better to run the QP several times. For example, for each unique X value, run QP on Y values of corresponding data.

Can you please clarify more on this? Are you suggesting to have a "group" parameter for the users?  so labels only affect other labels that they might realistically collide with (grouped by their Y positions when alignment="horizontal") ?

suhaani-agarwal:
To address the issue of non-colliding labels affecting each other's positions unnecessarily, I modified the `optimiseAlignedLabels()` function to first group labels based on their fixed coordinate (X for vertical alignment, Y for horizontal alignment) using precise bounding box calculations. This ensures labels that cannot possibly overlap are kept in separate groups and optimized independently. For horizontal alignment cases where labels share the same X position but have different Y positions (and thus don't collide), they now maintain their original positions without influencing each other, while still guaranteeing complete collision avoidance for labels that do overlap. 

This change implements your suggestion to run QP separately for groups of labels .
The solution now handles both vertical and horizontal alignment cases optimally, ensuring labels only affect others they might realistically overlap with.

Please let me know if any further modifications are required in this PR

tdhock:
good progress! Two comments

1. the duration option is not respected. can this be fixed please? I expected these labels to have smooth transitions, if they are specified. Can you please add a test case that fails for the current code, based on inst/examples/WorldBank-facets-map.R ? For an example of a test case involving smooth transitions, please see 
```
test-renderer2-widerect.R:406:test_that("middle of transition != after when duration=2000", {
```

2. I'm wondering if it is possible to hide the rectangular border around the text? Currently it is always shown with the same color as the text, right? do you think it is possible to do something like `geom_label_aligned(border_color=NA)` or should we make another geom, for example `geom_text_aligned()` which does not draw the border?

tdhock:
another thought: is it possible to implement the box as a polygon pointing to the data point? (instead of a rectangle around the text with no pointer) or is that too complicated? the code for drawing polygons with pointers in R is: 
```r
draw.polygons <- function(d,...){
  for(side in c("left", "right", "top", "bottom")){
    for(xy in c("x", "y")){
      col.name <- paste0(side, ".", xy)
      if(!col.name %in% names(d)){
        d[[col.name]] <- NA
      }
    }
  }
  if(! "box.color" %in% names(d)){
    d$box.color <- "black"
  }
  if(! "text.color" %in% names(d)){
    d$text.color <- "white"
  }
  for(i in 1:nrow(d))with(d[i,], {
    L <- list(
      x=c(left.x, left, top.x, right, right.x, right, bottom.x, left),
      y=c(left.y, top, top.y, top, right.y, bottom, bottom.y, bottom))
    for(xy.name in names(L)){
      xy <- L[[xy.name]]
      L[[xy.name]] <- xy[!is.na(xy)]
    }
    grid::grid.polygon(
      L$x, L$y,
      default.units="cm",
      gp=grid::gpar(col=box.color, fill=colour),
      name="directlabels.draw.polygon"
    )
  })
  d$colour <- d$text.color
  d
}
```

tdhock:
I added a new data set, data/parallelPeaks.RData, which illustrates an issue with the current code, shown below.
```r
gg_wrap <- ggplot()+
  theme_bw()+
  geom_segment(aes(
    start.seconds, process_i,
    xend=end.seconds, yend=process_i),
    data=parallelPeaks)+
  geom_point(aes(
    start.seconds, process_i),
    shape=1,
    data=parallelPeaks)+
  geom_blank(aes(
    start.seconds, process_i-1),
    shape=1,
    data=parallelPeaks)+
  geom_label_aligned(aes(
    start.seconds, process_i-0.5, label=peaks),
    alignment = "horizontal",
    data=parallelPeaks)+
  facet_wrap( ~ design + Fun, labeller=label_both, ncol=1)+
  scale_y_continuous(
    breaks=c(1,5,8,14))+
  scale_x_continuous("Time from start of computation (seconds)")
animint(gg_wrap+theme_animint(width=1500, height=1000))
```
code above has gives output below:
image

* top/first panel is OK
* second and third panels are not OK. I specified geom_label_aligned(alignment="horizontal") which means, for each unique Y value , run QP on X values. but we see some groups that result from running the QP with different Y values. For example panel 2, labels 98,82,89 are in the same group, but 98,89 have the smallest Y value, and 82 has a larger Y value. can you please create a test case based on this example which fails for the current code, and then push a fix?

note that the above code has height=1000 whereas we can run the same ggplot with height=2000 as below, and we get a different result.
```r
animint(gg_wrap+theme_animint(width=1500, height=2000))
```
code above gives result below:
image
the result above is expected. the result for height=1000 should be similar (but with labels with different Y values that overlap because there is not enough vertical space). 

Does that make sense?

suhaani-agarwal:
> 2\. I'm wondering if it is possible to hide the rectangular border around the text? Currently it is always shown with the same color as the text, right? do you think it is possible to do something like `geom_label_aligned(border_color=NA)` or should we make another geom, for example `geom_text_aligned()` which does not draw the border?

so currently, the text is drawn on a rect, the rect border is the same color as the text and corresponds to what the user specifies in the `color` parameter for geom_label_aligned. the color inside the rect corresponds to what the user specifies in the `fill` parameter for geom_label_aligned. 

I have now also added an option of **`background_rect = FALSE/TRUE`** . By this, the user can specify if the rect in the background of each label (including border) should be drawn or not. By default, this is set as TRUE. 

An example of this is as follows : In the `worldbank-facets-map.R` , I have added **`background_rect = FALSE`** : 
image

example of geom_label_aligned with **fill parameter:**
image

suhaani-agarwal:
> the result above is almost expected.
> 
> * in particular we see that 98,89 do not overlap, and 82 is above it (not involved with 98,89 position calculation), which is expected.

> the result for height=1000 should be similar (but with labels with different Y values that overlap because there is not enough vertical space).
> 
> Does that make sense?

Hi! currently, the grouping logic groups labels even if they are on **similar** y values (for horizontal alignment example) not **exactly** the same y values , so that the perpendicular overlapping can be avoided.
Currently, this "similar" y position accounts for the height of each label. If the labels are overlapping even a little in the perpendicular (y) axes, they will come under the same group and their X position will be optimised. 
 Because if we make the groups by exactly the same y values, something like below might happen :
image
Here alignment = "horizontal"
the labels are being aligned optimally horizontally only when they are on EXACTLY the same Y value and since there are no labels lying exactly on the same y value , none are grouped together and qp is not applied on any group. 

Although, the PredictedPeaks visualization will look as expected, the overlaps will be there in the axes perpendicular to the alignment exes (here, Y) 
image

Is that what is expected and should happen? Then geom_label_aligned will only be useful when we have labels on exactly the same x or y positions needing collision avoidance in only one axes.

suhaani-agarwal:
We can also consider adjusting the grouping logic to be less strict than it currently is. Right now, labels are grouped together if they overlap even slightly on the perpendicular axis—which is determined using the full height (for horizontal alignment) or full width (for vertical alignment) of the label. This can sometimes lead to labels being grouped and optimized together on alignment axis even when they are only slightly overlapping or close. [which is the issue right now that you pointed out in PredictedPeaks viz]

Instead, we could reduce this grouping threshold—so that labels are grouped only when they are very close in the perpendicular direction, not just slightly overlapping. This would still allow labels with nearly the same y positions (in horizontal alignment) to be grouped and optimized. 

image

At the same time, plots like PredictedPeaks, the output would remain unchanged and as expected because here the labels are only slightly overlapping and all labels are still distinctly visible.
image
This adjustment helps prevent unnecessary grouping of labels that are only slightly overlapping in the perpendicular direction, while still correctly optimizing labels that are visually close and could collide.

### Which approach should be preffered : 
 - Exact-position grouping
Group labels only if they lie on the exact same position along the axis perpendicular to alignment (e.g., same y for horizontal alignment, same x for vertical). This avoids overlap along the alignment axis but may allow overlaps on the perpendicular axis even if labels are slightly misaligned.

 - Similar-position grouping (with reduced threshold)
Group labels that lie on very close positions along the perpendicular axis. The similarity threshold is smaller than the full label size, so only slightly misaligned labels are grouped. Helps avoid major perpendicular overlaps without over-grouping.

suhaani-agarwal:
> * we see there is no rectangle drawn around label 0, but there is a rectangle drawn around all the other labels, which is not expected. 

The issue where label = 0 was incorrectly treated as null in measureText() (resulting in a rect with zero height and width) has been fixed.
image

I have also added test case to verify that a rect is rendered with non-zero width and height for label = 0.

tests/testthat/test-renderer1-geom-label-aligned.R
```diff
@@ -0,0 +1,399 @@
...
+
+library(animint2)
+data(WorldBank, package = "animint2")
+
+# subset of countries
+tracked_countries <- c("United States", "Vietnam", "India", "China", "Brazil",
+                      "Nigeria", "Mali", "South Africa", "Canada")
+
+# Filter WorldBank data
+wb <- WorldBank %>%
```
tdhock, line 11: please use data.table instead of dplyr (avoid adding another import)

tests/testthat/test-renderer1-geom-label-aligned.R
```diff
@@ -0,0 +1,376 @@
...
+    rect <- getNodeSet(group, './/rect')
+    expect_equal(length(rect), 1)
+    text <- getNodeSet(group, './/text')
+    expect_equal(length(text), 1)
+  }
+})
+
+test_that("label text content is correct", {
+  box_groups <- getNodeSet(info$html, '//g[@class="geom4_labelaligned_worldbankAnim"]//g[@class="geom"]')
+  actual_texts <- sapply(box_groups, function(group) {
```
tdhock, line 93: use getTextValue() in help-functions.R instead?

inst/examples/geom_label_aligned_examples.R
```diff
@@ -0,0 +1,137 @@
...
+      data = wb,
+      aes(x = fertility.rate, y = life.expectancy, color = group, key = country),
+      size = 8,
+      showSelected = "year",
+      clickSelects = "country"
+    ) +
+    geom_label_aligned(
+      data = label_data,
+      aes(x = fertility.rate, y = life.expectancy, label = label, fill = group, key = country),
+      alignment = "vertical", color = "#ffffd1", label_r = "9",
```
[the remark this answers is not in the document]
suhaani-agarwal: Done! I have made the input expectation for label_r numeric

tests/testthat/test-renderer1-geom-label-aligned.R
```diff
@@ -198,6 +195,54 @@ test_that("labels have at least 3px vertical spacing", {
...
+    xlab("Year") +
+    ylab("Life Expectancy (years)"),
+
+  first = list(),
+  selector.types = list(country = "multiple")
+)
+
+info <- animint2HTML(viz_dup)
+
+test_that("labels have at least 3px vertical spacing", {
```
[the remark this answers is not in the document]
suhaani-agarwal: thanks! Since the test is intentionally failing to show what happens with the default min_distance = 0.1, should I remove it now or update it to instead test that gaps can be less than 3px (but ≥ 0.1) here?

inst/htmljs/animint.js
```diff
@@ -1546,12 +1556,12 @@ var animint = function (to_select, json_file) {
                 })
                 .attr("y", function(d) {
                   if (alignment == "vertical") {
-                    return d.optimizedPos + (d.fontsize / 3);
+                    return d.optimizedPos + (d.size / 3);
                   } else {
-                    return d.scaledY - d.boxHeight * get_vjust(d) + d.boxHeight / 2 + ((d.fontsize || default_textSize) / 3);
+                    return d.scaledY - d.boxHeight * get_vjust(d) + d.boxHeight / 2 + (d.size / 3);
```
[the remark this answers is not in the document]
suhaani-agarwal: the d.size/3 adjustment helped vertically center text in labels by compensating for SVG's baseline positioning. Instead of this manual offset, I'm now using `.attr("dominant-baseline", "middle")` which handles text centering natively. This is a more reliable approach
I have removed `+ (d.size / 3)` now.
Thanks for pointing this out!

suhaani-agarwal:
> I added an expectation about font-size that fails using the current code on the World Bank test case.
> geom_label_aligned(size=5) has no effect (font-size: 12px is used no matter what the user specifies as size param). can you please fix?

This issue has been fixed now, the user will now be able to specify the size parameter for fontsize of the text. 
for example for geom_label_aligned(size = 5) in the same plot, we now get:

image

tests/testthat/test-renderer1-geom-label-aligned.R
```diff
@@ -0,0 +1,376 @@
...
+# ─────────────────────────────────────────────────────────────────────────────
+# Test: Verify that the font size of Medium group labels (labels close to each other in this case) has decreased
+# after adding more crowded labels, indicating the shrinking mechanism is working correctly.
+test_that("Font size shrinks for Medium labels after crowding (vs. initial)", {
+  # XPath to target the <text> elements inside <g class="geom" id="MediumXXX">
+  medium_label_text_xpath <- '//g[@class="geom2_labelaligned_orangeGrowth"]//g[starts-with(@id, "Medium")]//text'
+  initial_text_nodes <- getNodeSet(info$html, medium_label_text_xpath)
+  initial_font_sizes_num <- sapply(initial_text_nodes, function(node) {
+    as.numeric(gsub("px", "", xmlGetAttr(node, "font-size")))
+  })
```
tdhock, line 343: can you please use getStyleValue instead? (if it works? I think it should)
```r
  ts_size <- getStyleValue(info$html, '//g[@class="geom2_labelaligned_lifeExpectancyPlot"]//text', "font-size")
```

inst/htmljs/animint.js
```diff
@@ -1429,6 +1459,117 @@ var animint = function (to_select, json_file) {
...
+                      return d.scaledX - d.boxWidth * get_hjust(d);
+                    } else {
+                      return d.optimizedPos - d.boxWidth / 2;
+                    }
+                  })
+                  .attr("y", function(d) {
+                    if (alignment == "vertical") {
+                      return d.optimizedPos - d.boxHeight / 2;
+                    } else {
+                      return d.scaledY - d.boxHeight * get_vjust(d);
```
tdhock, line 1536: vjust is not supported anywhere else in the code -- please add test cases for values other than 0.5

R/geom-label-aligned.R
```diff
@@ -0,0 +1,155 @@
...
+#' - Boxes are aligned along the horizontal axis at their original Y positions
+#'
+#' The QP solver minimizes the total squared distance from original positions while
+#' enforcing minimum spacing constraints between boxes.
+#'
+#' @inheritParams layer
+#' @inheritParams geom_point
+#' @param label_r Radius of rounded corners. Defaults to 0.15 lines.
+#' @param alignment One of "vertical" (QP on Y axis) or "horizontal" (QP on X axis)
+#' @param min_distance Minimum distance between boxes in native units.
```
tdhock, line 24: please clarify units for min_distance.
it is written native units, which I interpret to mean the units displayed on the axis.
However I tried size=40 with min_distance=5 and the distance between boxes is clearly smaller than 5 axis units.
image

tests/testthat/test-renderer1-geom-label-aligned.R
```diff
@@ -0,0 +1,376 @@
...
+})
+
+test_that("label_r sets correct rx and ry values", {
+  rect_node <- getNodeSet(
+    info$html,
+    '//g[@class="geom4_labelaligned_worldbankAnim"]//g[@class="PANEL1"]//g[@class="geom"]//rect'
+  )[[1]]
+  attrs <- xmlAttrs(rect_node)
+  expect_equal(as.numeric(attrs[["rx"]]), 5)
+  expect_equal(as.numeric(attrs[["ry"]]), 5)
```
tdhock, line 183: why is rx 5 if label_r=9 ?

tdhock:
This is really excellent work! Please fix the few things I mentioned before I merge. We are almost done.

tests/testthat/test-renderer1-geom-label-aligned.R
suhaani-agarwal: I’ve now updated the tests to use `getPropertyValue` because in geom_label_aligned, the font-size is directly set as an attribute on the `<text>` element, not inside the style. So getStyleValue() wouldn’t retrieve it correctly in this case.

tests/testthat/test-renderer1-geom-label-aligned.R
suhaani-agarwal: The original label_r was set to 5, which is why the rx/ry values were tested against 5 and were passing. It seems the value got updated to 9 in a later commit. I have now reverted it back to 5.

R/geom-label-aligned.R
suhaani-agarwal: Thanks for pointing that out, the earlier comment was misleading. min_distance is actually in pixels, not in native/axis units. I’ve updated the documentation accordingly to reflect this.

inst/htmljs/animint.js
suhaani-agarwal: Done, I have added a test for vjust values 1, 0 and 0.5 for horizontal alignment.

inst/htmljs/animint.js
```diff
@@ -1429,6 +1459,117 @@ var animint = function (to_select, json_file) {
...
+        var textStyle = [
+            "font-family:" + (d.family || "sans-serif"),
+            "font-weight:" + (d.fontface == 2 ? "bold" : "normal"),
+            "font-style:" + (d.fontface == 3 ? "italic" : "normal")
+        ].join(";");
+        // Use d.size for font size (always current value)
+        var fontsize = d.size;
+        var textSize = measureText(d.label, fontsize, d.angle, textStyle);
+        d.boxWidth = textSize.width;
+        d.boxHeight = textSize.height;
```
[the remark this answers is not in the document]
suhaani-agarwal: That's because getBBox() returns the bounding box of the text, which includes space below the baseline to account for descenders (like the letter 'g') and general font metrics.

For example, in the image below, Nigeria appears visually centered due to the descender in 'g', while South Africa, which lacks descenders, appears slightly higher despite having the same alignment:

image

We could experiment with a small manual offset for y position of `<text>` to better visually center the text--like adjusting the vertical position using + d.size / 3, which I had tried earlier. Let me know if you'd like me to explore that again

suhaani-agarwal:
> Also there is a problem with the smooth transitions. When the geom appears, it flies from the upper left of the plot, but it should just appear instantly (no smooth transition).
> See video:

Hi, I have updated the code to fix this issue, the labels appear instantly and dont fly in from the upper left anymore

tests/testthat/test-renderer1-geom-label-aligned.R
```diff
@@ -175,22 +175,17 @@ test_that("Aligned labels in timeSeries do not collide after selection/deselecti
...
-  rect_node <- getNodeSet(
-    info$html,
-    '//g[@class="geom4_labelaligned_worldbankAnim"]//g[@class="PANEL1"]//g[@class="geom"]//rect'
-  )[[1]]
-  attrs <- xmlAttrs(rect_node)
-  expect_equal(as.numeric(attrs[["rx"]]), 5)
-  expect_equal(as.numeric(attrs[["ry"]]), 5)
+  rx <- getPropertyValue(info$html, '//g[@class="geom4_labelaligned_worldbankAnim"]//rect', "rx")
+  ry <- getPropertyValue(info$html, '//g[@class="geom4_labelaligned_worldbankAnim"]//rect', "ry")
+  expect_true(all(as.numeric(rx) == 5))
```
tdhock, line 180: please use expect_equal

suhaani-agarwal: done!

inst/htmljs/animint.js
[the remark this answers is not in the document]
tdhock: ok, that makes sense. 
it does look a bit strange for numbers which lack the descenders, but I think that we should keep this logic for now, for simplicity.

tdhock:
I did some interactive testing and the new features seem to be working well.

tdhock:
there seems to be an issue with vjust.
typically vjust=0 means bottom alignment, and vjust=1 means top alignment.
for example consider this
```r
df <- data.frame(vjust=c(0,0.5,1))
ggplot(df)+
  geom_point(aes(vjust,vjust), color="red")+
  geom_text(aes(vjust,vjust,label=vjust,vjust=vjust), size=5)
```
image

but to get top alignment in the upper left world bank time series data viz as below
image
I had to just `vjust=0` which is opposite of expected. 
Can you please add a test case which fails for the current code?
And then push a fix?

(by the way this vertical and horizontal label functionality is really cool!!)

tests/testthat/test-renderer1-geom-label-aligned.R
```diff
@@ -177,8 +177,8 @@ test_that("Aligned labels in timeSeries do not collide after selection/deselecti
 test_that("label_r sets correct rx and ry values", {
   rx <- getPropertyValue(info$html, '//g[@class="geom4_labelaligned_worldbankAnim"]//rect', "rx")
   ry <- getPropertyValue(info$html, '//g[@class="geom4_labelaligned_worldbankAnim"]//rect', "ry")
-  expect_true(all(as.numeric(rx) == 5))
-  expect_true(all(as.numeric(ry) == 5))
+  expect_equal(as.numeric(rx), rep(5, length(rx)))
```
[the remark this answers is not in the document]
suhaani-agarwal: done!

tests/testthat.R
```diff
@@ -7,31 +7,24 @@ source("helper-HTML.R")
...
-} else {
-  # Non-browser tests
-  tests_run(filter=filter)
-}
\ No newline at end of file
+}
+# Run tests
+tests_run(filter=filter)
+# Save coverage and cleanup
+if(coverage_active) {
```
tdhock, line 27: I see this error in compiler tests
```
── Error ('test-compiler-ghpages.R:42:3'): animint2pages() returns list of meta-data ──
<error_stop_port_search/rlang_error/error/condition>
Error in `startup(port = port, ...)`: Chrome debugging port not open after 10 seconds.
Backtrace:
     ▆
  1. └─animint2::animint2pages(...) at test-compiler-ghpages.R:42:3
  2.   └─chromote::ChromoteSession$new()
  3.     └─chromote (local) initialize(...)
  4.       └─chromote::default_chromote_object()
  5.         ├─chromote::set_default_chromote_object(Chromote$new())
  6.         └─Chromote$new()
  7.           └─chromote (local) initialize(...)
  8.             └─Chrome$new()
  9.               └─chromote (local) initialize(...)
 10.                 └─chromote:::launch_chrome(path, args)
 11.                   └─chromote:::with_random_port(...)
 12.                     ├─base::tryCatch(...)
 13.                     │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
 14.                     │   └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
 15.                     │     └─base (local) doTryCatch(return(expr), name, parentenv, handler)
 16.                     └─chromote (local) startup(port = port, ...)
 17.                       └─rlang::abort(...)
 ```
which seems to indicate that chromote is required but not available for the gh pages compiler test.
I believe tests_init() starts chromote, so here I remove the if statement, so tests_init() will be run in the compiler tests.

Comment (tdhock) [reply tests/testthat.R:27]: I'm not sure why we were not getting this error before this PR?

**Comment by tdhock** [comment]
this is excellent work! I updated the world bank map data viz https://tdhock.github.io/2025-01-WorldBank-facets-map/ to use `geom_label_aligned`
it is really great how the text labels automatically decrease in size when there are too many of them to fit in the available space at the original size, as below.
image

File: tests/testthat/test-renderer1-geom-label-aligned.R [hunk 21]
```diff
@@ -0,0 +1,385 @@
...
+      showSelected = "Tree",
+      clickSelects = "Tree"
+    ) +
+    scale_color_manual(
+      values = c(Fast = "#E41A1C", Medium = "#377EB8", Slow = "#4DAF4A"),
+      name = "Growth Rate"
+    ) +
+    scale_fill_manual(
+      values = c(Fast = "#E41A1C", Medium = "#377EB8", Slow = "#4DAF4A"),
+      name = "Growth Rate"
```
tdhock, line 243: please refactor to an intermediate variable to avoid repeating these values

tests/testthat.R
suhaani-agarwal: Hi! actually we were getting this error before this PR and the coverage PR too, see : https://github.com/animint/animint2/actions/runs/15376944905/job/43262818512
This error automatically went away after some commits. And we were getting this error randomly in some commits like : https://github.com/animint/animint2/actions/runs/15497830316/job/43638653143 and more..
This error appears randomly sometimes and sometimes it doesnt. 

Im not sure but its maybe because the Chromote session is failing to start properly within the timeout period sometimes?

tdhock: ok. so it was appearing before but randomly.
the fix I pushed seemed to have fixed it, but maybe that is random?

tests/testthat/test-renderer1-geom-label-aligned.R
suhaani-agarwal: done!

tdhock:
great, thanks for all your hard work on this very excellent new feature!!

Issue #210, opened 2025-07-23
coverage documentation


tdhock:
hi @suhaani-agarwal can you please add some documentation on the wiki about how the current R and JS coverage works?
I see the following
https://github.com/animint/animint2/actions/runs/16479846665/job/46590978837
https://github.com/animint/animint2/blob/master/.github/workflows/tests.yaml
<img width="1873" height="958" alt="Image" src="https://github.com/user-attachments/assets/1bb2e5af-664b-496d-9ecd-e868cb9ef5c2" />
it says on the right actions log: coverage saved to a json file, coverage reports upload successfully. This indicates to me that the JS coverage was computed and uploaded, is that right?

After that, there are two actions steps, Convert JS coverage to Istanbul format, and Upload JS coverage to Codecov, which appear to have been skipped. This seems to contradict the previous output that says JS coverage was successfully uploaded. Are there two different uploads required? can you please clarify how this works on a wiki page? (create a new one called coverage?)

suhaani-agarwal:
Hi! I have added 2 wiki pages:  
1) https://github.com/animint/animint2/wiki/Coverage-Collection-in-GitHub-Actions
2) https://github.com/animint/animint2/wiki/Understanding-and-Using-Codecov 

Is this okay?

tdhock:
I think this is false, can you please double check or clarify?

> The other job runs the same tests under covr::package_coverage(), which collects R coverage across all test files, including the renderer tests, without needing Chromote

renderer tests do need chromote.

suhaani-agarwal:
> I think this is false, can you please double check or clarify?
> 
> > The other job runs the same tests under covr::package_coverage(), which collects R coverage across all test files, including the renderer tests, without needing Chromote
> 
> renderer tests do need chromote.

You're right , thanks for pointing that out!
What I meant was that while Chromote does get initialized there, JavaScript coverage isn’t collected, likely due to how covr::package_coverage() runs the tests in a temporary instrumented directory,
I have updated that part now, please check

tdhock:
great thanks

PR #209, opened 2025-07-07
shiny tests using chromote

biplab-sutradhar:
Closes #143 

This PR migrates RSelenium tests to Chromote for animint2, covering examples/shiny, examples/shiny-WorldBank, and examples/rmarkdown visualizations.

PR #212, opened 2025-07-24
Combining renderer and compiler tests

### LINKED ISSUE #211: combine compiler and renderer tests?
we may attempt to combine all non-CRAN tests (currently run in separate compiler and renderer CI jobs).

* CRAN would not start chromote, nor JS coverage.
* combined compiler/renderer job would start chromote, and compute R/JS coverage.

Related to #210 

what do you think @suhaani-agarwal ?

suhaani-agarwal:
Closes #211 

Related to #210


tdhock:
do you think it's going to work and be beneficial?
I am a bit hesitant when I see the GitHub action time of 30 minutes.
how much slower is that than the current set up?

tests/testthat.R
```diff
@@ -5,26 +5,25 @@ setwd("testthat")
...
   if(coverage_active) {
     message("JS coverage collection started")
   }
 }
 # Run tests
-tests_run(filter=filter)
+message("\n=== Running COMPILER tests ===")
+tests_run(filter = "compiler")
+message("\n=== Running RENDERER tests ===")
+tests_run(filter = "renderer")
```
tdhock, line 24: sometimes interactively we want to set filter to something else.
so I wonder if we need to separate these two?
why not tests_run() (without filter arg) ?

suhaani-agarwal: I had separated the two mainly to make the workflow logs a bit clearer , so it's easy to see when compiler-based tests are running and when they switch to renderer. But if you’d prefer running them all together without that printed separation, I can definitely update it, would you like me to do that?

suhaani-agarwal: running tests_run() without a filter would also run two other test files that I believe are currently not in working condition and produce errors:

 - [test-shiny.R](https://github.com/animint/animint2/blob/master/tests/testthat/test-shiny.R), which is being updated in PR #209 and
 - [test-tt-inter.R](https://github.com/animint/animint2/blob/master/tests/testthat/test-tt-inter.R), which seems outdated now that the new tooltip implementation works and [test-renderer3-tooltip-interactivity.R](https://github.com/animint/animint2/blob/master/tests/testthat/test-renderer3-tooltip-interactivity.R) covers this test well.

Do you think we should remove test-tt-inter.R?

Once these test files are free of errors, we can update this part of the code to use tests_run() without the separate renderer and compiler filters, if you prefer that approach instead of printing the separation like it currently does.

tdhock: yes please remove test-tt-inter.R

tdhock: about the shiny tests, can you please review https://github.com/animint/animint2/pull/209 ?

tdhock:
these changes seem OK overall.
but why do we need two separate CI jobs for JS and R coverage? Is it possible to do both JS and R coverage in the same CI job? if not, why not?

tests/testthat.R
suhaani-agarwal: > about the shiny tests, can you please review #209 ?

Yes, I’ve gone through the PR and the related issue and dropped some comments about optimising it and aligning it better with the existing tests. Please take a look when you're free , and let me know if I’ve misunderstood anything or if anything needs clarification!

tdhock:
I like it! Thanks for the clear explanation. Can you please copy that explanation to https://github.com/animint/animint2/wiki/Testing too? (or another page if you think that is better)
That would close #210

PR #214, opened 2025-07-31
Fixing geom abline bug

### LINKED ISSUE #142: geom_abline go out of plot range
for example https://github.com/tdhock/changepoint-data-structure/blob/master/figure-loss-small-data.R makes
image

suhaani-agarwal:
Resolves #142

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1357,20 +1357,58 @@ var animint = function (to_select, json_file) {
       if (g_info.geom == "segment") {
 	g_info.style_list = line_style_list;
 	eActions = function (e) {
-          e.attr("x1", function (d) {
-            return scales.x(d["x"]);
-          })
-            .attr("x2", function (d) {
-              return scales.x(d["xend"]);
-            })
-            .attr("y1", function (d) {
-              return scales.y(d["y"]);
-            })
-            .attr("y2", function (d) {
-              return scales.y(d["yend"]);
-            })
-	};
-	eAppend = "line";
+    e.each(function(d) {
+      var line = d3.select(this);
+      if (g_info.is_abline) {
+        var slope = g_info.abline_params.slope;
+        var intercept = g_info.abline_params.intercept;
+        var xDomain = scales.x.domain();
+        var yDomain = scales.y.domain();
+        // calculate endpoints
+        var x1 = xDomain[0];
+        var y1 = slope * x1 + intercept;
+        var x2 = xDomain[1];
+        var y2 = slope * x2 + intercept;
+        // Clip to y domain
+        if (y1 < yDomain[0]) {
+          x1 = (yDomain[0] - intercept) / slope;
+          y1 = yDomain[0];
+        } else if (y1 > yDomain[1]) {
+          x1 = (yDomain[1] - intercept) / slope;
+          y1 = yDomain[1];
+        }
+        if (y2 < yDomain[0]) {
+          x2 = (yDomain[0] - intercept) / slope;
+          y2 = yDomain[0];
+        } else if (y2 > yDomain[1]) {
+          x2 = (yDomain[1] - intercept) / slope;
+          y2 = yDomain[1];
+        }
+        // Handle vertical lines (infinite slope)
+        if (!isFinite(slope)) {
+          x1 = x2 = intercept;
+          y1 = yDomain[0];
+          y2 = yDomain[1];
+        }
+        // Handle horizontal lines (zero slope)
+        if (slope === 0) {
+          y1 = y2 = intercept;
+          x1 = xDomain[0];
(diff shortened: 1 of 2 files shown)


suhaani-agarwal:
earlier:
image

now:
image

suhaani-agarwal:
As discussed in the issue, the ablines were extending beyond the plot boundaries only when update_axes was applied. This happened because the lines were being clipped on the R (compiler) side, but not re-clipped on the JavaScript (renderer) side after the axes were updated. Since update_axes can change the view limits , the original clipping logic in R side failed to keep the ablines within bounds after update_axes was called.

To fix this, I have now moved the clipping logic and the calculation of x, xend, y, and yend from the compiler to the renderer side, so that the lines are always clipped correctly based on the current axes after any update. The lines are now correctly being clipped as shown in the image even when update_axes is applied.

However, after this change, there seem to be differences in the coordinates plotted for the ablines compared to before. The visual result is not exactly the same, and it may be due to how the renderer is now computing these positions.
I'll investigate this further to understand the discrepancy

I’ve also added a test case to verify that ablines remain clipped within the plot boundaries even after update_axes is applied.

inst/htmljs/animint.js
```diff
@@ -1355,13 +1355,13 @@ var animint = function (to_select, json_file) {
       }
       data_to_bind = data;
       if (g_info.geom == "segment") {
-	g_info.style_list = line_style_list;
-	eActions = function (e) {
-    e.each(function(d) {
+  g_info.style_list = line_style_list;
+  eActions = function (e) {
+    e.each(function(d, i) {
```
tdhock, line 1360: I don't think you should need each here... can't a data-join be used instead?

suhaani-agarwal: I have updated the code to use data-join instead now

suhaani-agarwal:
Hi @tdhock , please review this PR, I have fixed the geom_abline bug

inst/htmljs/animint.js
```diff
@@ -1355,22 +1355,54 @@ var animint = function (to_select, json_file) {
...
+      })
+      .attr("x2", function(d) {
+        var slope = d.slope !== undefined ? +d.slope : (g_info.abline_params && g_info.abline_params.slopes) ? g_info.abline_params.slopes[d.i] : 1;
+        var intercept = d.intercept !== undefined ? +d.intercept : (g_info.abline_params && g_info.abline_params.intercepts) ? g_info.abline_params.intercepts[d.i] : 0;
+        var xDomain = scales.x.domain();
+        var yDomain = scales.y.domain();
+        var x2 = xDomain[1];
+        var y2 = slope * x2 + intercept;
+        if (y2 < yDomain[0]) x2 = (yDomain[0] - intercept) / slope;
+        if (y2 > yDomain[1]) x2 = (yDomain[1] - intercept) / slope;
```
tdhock, line 1380: there is duplicated logic here. can you please refactor to clarify your intent?

tests/testthat/test-renderer1-geom-abline.R
```diff
@@ -75,3 +77,70 @@ abline.at.5 <- slope*(cxy$cx-abl$x1)+abl$y1
...
+  if(length(visible_lines) > 0) {
+    # Extract coordinates of visible lines
+    coords <- t(sapply(visible_lines, function(line) {
+      attrs <- xmlAttrs(line)
+      c(x1 = as.numeric(attrs["x1"]),
+        x2 = as.numeric(attrs["x2"]),
+        y1 = as.numeric(attrs["y1"]),
+        y2 = as.numeric(attrs["y2"]))
+    }))
+    test_that("visible lines are properly clipped", {
```
tdhock, line 135: please avoid test_that block inside another test_that block

inst/htmljs/animint.js
suhaani-agarwal: done!

tests/testthat/test-renderer1-geom-abline.R
```diff
@@ -75,3 +72,56 @@ abline.at.5 <- slope*(cxy$cx-abl$x1)+abl$y1
...
+  cat(sprintf("\nVisible lines: %d/%d\n", length(visible_lines), length(ablines)))
+  expect_gt(length(visible_lines), 0)
+})
+
+test_that("visible lines are properly clipped within plot area", {
+  if (length(visible_lines) > 0) {
+    coords <- t(sapply(visible_lines, get_line_coords))
+    expect_true(all(coords >= 0 & coords <= 400), info = "All visible line coordinates should be within the plot area")
+  } else {
+    skip("No visible lines found for clipping test")
```
tdhock, line 125: is this skip branch necessary because of the runif randomness?

could you please refactor to use a known number of visible lines, to avoid this if else?

Issue #219, opened 2025-08-31
tooltip="two\nlines" causes data to be not rendered


tdhock:
Hi @suhaani-agarwal 
I tried to use `aes(tooltip)` with a color legend and with line breaks (`\n`) in the text string, and I expected this should work (or at least produce an error/warning indicating that line breaks are not supported).
However I observed that when there is both a line break and a color legend, the corresponding element disappears, which is not expected.
For example consider this code
```r
animint(
  ggplot()+
    geom_point(aes(
      x, x, tooltip=x, color=x),
      size=5,
      data.frame(x=c("one line", "two\nlines"))))
```
On my machine I get the result below

<img width="593" height="926" alt="Image" src="https://github.com/user-attachments/assets/bfc615b0-77fc-4e93-85b3-c94ad82c4a4a" />

above we see that there is a data point for "one line" but not for "two\nlines" which is un-expected (we should see both data points).

For comparison, if we remove `aes(color)` we see that the point renders, as in the code below
```r
animint(
  ggplot()+
    geom_point(aes(
      x, x, tooltip=x),
      size=5,
      data.frame(x=c("one line", "two\nlines"))))
```
<img width="594" height="497" alt="Image" src="https://github.com/user-attachments/assets/face7458-77fb-43d7-96e3-02141efb014d" />

on my machine the tooltip says "two" not "two\nlines" which is better than not rendering any data point, but in this case I would have expected a warning which says like "aes(tooltip) does not support line breaks; please remove \n from aes(tooltip)"

Can you please investigate and fix?

suhaani-agarwal:
Hi! I will investigate into this issue further!

PR #220, opened 2025-09-01
Fixing tooltip='two\nlines' data not rendering issue

suhaani-agarwal:
fixing #219

### DIFF
--- a/R/z_animintHelpers.R
+++ b/R/z_animintHelpers.R
@@ -980,7 +980,9 @@ saveChunks <- function(x, meta){
   if(is.data.frame(x)){
     this.i <- meta$chunk.i
     csv.name <- sprintf("%s_chunk%d.tsv", meta$g$classed, this.i)
-    data.table::fwrite(x, file.path(meta$out.dir, csv.name), quote=FALSE,
+  # Ensure fields are quoted so that embedded newlines or tabs in
+  # string fields do not break the TSV format when read by d3.tsv.
+    data.table::fwrite(x, file.path(meta$out.dir, csv.name), quote=TRUE,
                 row.names=FALSE, sep="\t")
     meta$chunk.i <- meta$chunk.i + 1L
     this.i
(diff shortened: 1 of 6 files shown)


suhaani-agarwal:
Now the newline survives serialization, but the browser’s default rendering (and/or the way the tooltip is inserted into the DOM) collapses it into a single space. This happens not because the code explicitly replaces \n with a space, but because HTML collapses newlines by default.

We have a few options:

 - Keep quote=TRUE (as I have now) and accept that \n renders as a space in tooltips.
 - Keep quote=TRUE and implement multi-line tooltips by converting \n to `<br>` for proper HTML rendering.
 - Revert to quote=FALSE and remove or warn about \n in tooltips like ('aes(tooltip) does not support line breaks; please remove \n from aes(tooltip)') on the R side to avoid TSV corruption.

Could you advise which approach we should go with?

R/z_animintHelpers.R
```diff
@@ -980,7 +980,9 @@ saveChunks <- function(x, meta){
   if(is.data.frame(x)){
     this.i <- meta$chunk.i
     csv.name <- sprintf("%s_chunk%d.tsv", meta$g$classed, this.i)
-    data.table::fwrite(x, file.path(meta$out.dir, csv.name), quote=FALSE,
+  # Ensure fields are quoted so that embedded newlines or tabs in
+  # string fields do not break the TSV format when read by d3.tsv.
```
tdhock, line 984: please fix indentation (# should be aligned with surrounding text, but it is before)

inst/htmljs/animint.js
```diff
@@ -1797,8 +1797,9 @@ var animint = function (to_select, json_file) {
         mouseX = d3.event.pageX;
         mouseY = d3.event.pageY;
       }
+      var safeHtml = String(content).replace(/\n/g, '<br/>');
```
tdhock, line 1800: if we do this on R instead of in JS, I think we shold be able to get line breaks in legend entries too, see test below.
image

maybe also desirable in

* axes titles
* axes tick marks
* facet titles
* geom_text label

but all of these features are too much for this PR, so I will create a new issue.

PR #257, opened 2025-10-19
Add linter to ensure no tabs in animint.js

### LINKED ISSUE #256: add linter to ensure no tabs in animint.js
tabs in animint.js can lead to confusing diffs and other display.

a CI job should be added to fail if there are any tabs added.

ANAMASGARD:
This PR adds two checks to prevent tabs from being introduced into `inst/htmljs/animint.js`:

1. **GitHub Actions workflow** (`.github/workflows/check-tabs.yml`) - Fails CI if tabs are detected
2. **R test** (`tests/testthat/test-linter-no-tabs.R`) - Tests during package checks

## Testing

Tested locally by:

1. **Verified no tabs in source:**
   ```bash
   grep -P '\t' inst/htmljs/animint.js
   # No output - no tabs found ✓
   ```

2. **Installed package and ran test (should pass):**
   ```bash
   R CMD INSTALL .
   Rscript -e 'library(testthat); library(animint2); test_file("tests/testthat/test-linter-no-tabs.R")'
   # Result: PASS (2 tests passed) ✓
   ```
  
screen-pic-1

3. **Added tab to line 10 and tested (should fail):**
   ```bash
   R CMD INSTALL .
   Rscript -e 'library(testthat); library(animint2); test_file("tests/testthat/test-linter-no-tabs.R")'
   # Result: FAIL - "Tab characters found in animint.js at lines: 10" ✓
   ```
 
pick-part-2

4. **Removed tab and verified test passes again:**
   The test correctly detects when tabs are present and passes when they're removed.

## Files Changed

- `.github/workflows/check-tabs.yml` - New CI workflow
- `tests/testthat/test-linter-no-tabs.R` - New R test

Fixes #256

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,3 +1,4 @@
+CONTRIBUTING.md
 readme_website.md
 \.github
 ^.*\.Rproj$


tests/testthat/test-linter-no-tabs.R
```diff
@@ -0,0 +1,20 @@
+test_that("animint.js contains no tab characters", {
+  animint_js_path <- system.file("htmljs", "animint.js", package = "animint2")
```
tdhock: please add mustWork=TRUE

.github/workflows/check-tabs.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
+jobs:
+  check-tabs:
+    runs-on: ubuntu-latest
+    name: Check animint.js for tabs
```
tdhock: it is a good idea to have linting be a separate CI job like this.
but it is redundant with the testthat file below.
please delete one or the other.
If you keep this file, can lintr package be used?
see data.table for an example: https://github.com/Rdatatable/data.table/blob/master/.github/workflows/code-quality.yaml

tdhock:
please avoid cross-posting (you wrote the same comment in #256 ).

can you please push a commit which adds a tab to animint.js, so we can see that the test would correctly fail in that case?
(and then push another commit which removes the tab)

ANAMASGARD:
**Root cause**: The test `test-compiler-ghpages.R` tries to DELETE a GitHub repository but lacks permissions:

❌ R_coverage fails due to pre-existing GitHub API permissions issue

Sir should I proceed with this PR as-is, or would you like me to investigate the ghpages test issue separately in a different PR?

tdhock:
great thanks

PR #271, opened 2025-10-28
Add warning for geom_point shape != 21

### LINKED ISSUE #270: geom_point() should warn for shape!=21
animint2 web rendering only supports shape=21 so if the user tries to do something else, we should emit a warning.

```r
geom_point(aes(shape=something))
geom_point(shape=1)
```

ANAMASGARD:
Fixes #270

## Summary
Implements warning when users specify shape values other than 21 in `geom_point()`, since animint2 web rendering only supports shape=21.

### ✅ First Commit: Failing Test (Current)
Added test that demonstrates the issue. **Test currently fails as expected**, proving that the warning is not yet implemented.

**Test Output Screenshot:**
Screenshot From 2025-10-28 19-32-26

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# Changes in version 2025.10.31 (PR#270)
+# Changes in version 2025.10.31 (PR#271)
 
 - `geom_point()` now warns when shape parameter is set to a value other than 21, since animint2 web rendering only supports shape=21 for proper display of both color and fill aesthetics.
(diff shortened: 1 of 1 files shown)


tdhock (changes requested):
thanks!!

tdhock (changes requested):
also can you please increase version in DESCRIPTION and add NEWS item?

PR #274, opened 2025-11-16
Fix axis tick font-size inconsistency with update_axes (Fix #273)

### LINKED ISSUE #273: axis tick text size inconsistent with update_axes
Image

above we can see the bottom plot (`update_axes="x"`)x axis tick text size is smaller than the plot above that (no `update_axes`).
this text size should be consistent.

ANAMASGARD:
FIXES #273 

I tried reproducing the issue locally using a simple test case with mtcars data:

**Screenshot showing the bug:**
Screenshot From 2025-11-16 20-25-34

- **Left plot**: No `update_axes` → tick labels render at correct size (12pt)
- **Right plot**: With `update_axes=c("x")` → tick labels render smaller (browser default ~9px)

Both plots explicitly set `theme(axis.text = element_text(size=12))`, but only the left plot respects this setting.

## Root Cause
The `update_axes()` function in `inst/htmljs/animint.js` (line ~2006) calls `.call(xyaxis)` but doesn't preserve the font-size style that was set during initial rendering.

## Next Steps What I will do 
In the next commit  :=> 
1. Fix `inst/htmljs/animint.js` to preserve font-size in `update_axes()` function
2. Verify the test passes
3. Confirm all existing tests still pass

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: animint2
 Title: Animated Interactive Grammar of Graphics
-Version: 2025.10.31
+Version: 2025.11.17
 URL: https://animint.github.io/animint2
 BugReports: https://github.com/animint/animint2/issues
 Authors@R: c(


tests/testthat/test-renderer3-update-axes-tick-font-size.R
```diff
@@ -0,0 +1,29 @@
...
+  with_update_xpath <- '//svg[@id="plot_withUpdateX"]//g[@class="xaxis axis xaxis_1"]//g[@class="tick major"]//text'
+  no_update_font_size <- getStyleValue(info$html, no_update_xpath, "font-size")
+  with_update_font_size <- getStyleValue(info$html, with_update_xpath, "font-size")
+  expect_gt(length(no_update_font_size), 0)
+  expect_gt(length(with_update_font_size), 0)
+  no_update_size_unique <- unique(no_update_font_size)
+  with_update_size_unique <- unique(with_update_font_size)
+  expect_equal(length(no_update_size_unique), 1)
+  expect_equal(length(with_update_size_unique), 1)
+  expect_equal(with_update_size_unique, no_update_size_unique, info=sprintf("Axis tick text font-size should be consistent regardless of update_axes. Expected %s but got %s", no_update_size_unique, with_update_size_unique))
```
tdhock: this test is good!

PR #303, opened 2026-03-01
add rosling bubbles test

aakritixyz:
i'm adding test-renderer-rosling-bubbles.R to verify interactive bubble plots.

local Verification:
my previous test video didn't utilize the internal Chromote functions. I have now refactored the environment to use a Chromote backend, allowing for a post-JS rendered DOM capture.

updated Video Demo: [https://youtu.be/F5YkaAXtUbo]

@suhaani-agarwal, @tdhock   please review the test logic. I have also patched the tests_init() helpers locally to support this Chromote workflow; let me know if those should be included here or in a separate PR.


aakritixyz:
@suhaani, thanks for the feedback!

i’ve updated the test to use the new parameter-based syntax for showSelected and clickSelects. i also extended it to validate interactive elements like the Play/Pause widgets and selection attributes.

PR #306, opened 2026-03-02
animint.js: remove unnecessary Selectors.hasOwnProperty checks

### LINKED ISSUE #278: animint.js Selectors does not need hasOwnProperty
there are several uses of Selectors.hasOwnProperty which seems to be un-necessary, so should be removed.
```
grep  -nH --null Selectors.hasOwnProperty animint.js
animint.js:1570:            if (Selectors.hasOwnProperty(selector_name)) {
animint.js:1866:    if(Selectors.hasOwnProperty(selector_name)){
animint.js:2087:    if(!Selectors.hasOwnProperty(v_name)){
animint.js:2766:        if(Selectors.hasOwnProperty(selector_nam)){
animint.js:2797:        }//if(Selectors.hasOwnProperty(selector_nam))

Grep finished with matches found at Tue Nov 25 09:55:05
```

ANAMASGARD:
Fixes #278 

- Replaced Selectors.hasOwnProperty(selector_name) with a simple selector_name check in draw_geom and geom_label_aligned. 
- Removed redundant guard in update_selector.
- Removed Selectors.hasOwnProperty check in update_axes.

I confirm I have not used any AI code generation tools for this PR.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: animint2
 Title: Animated Interactive Grammar of Graphics
-Version: 2026.2.28
+Version: 2026.3.2
 URL: https://animint.github.io/animint2
 BugReports: https://github.com/animint/animint2/issues
 Authors@R: c(
@@ -64,7 +64,7 @@ Authors@R: c(
      comment="Animint2 GSoC 2025"),
     person("Gaurav", "Chaudhary",
      role="ctb",
-     comment="Remove unused css.file parameter; fix issue #233 selector values; fix issue #273 update_axes tick font-size; fix issue #276 update_axes transition duration"))
+     comment="Remove unused css.file parameter; fix issue #233 selector values; fix issue #273 update_axes tick font-size; fix issue #276 update_axes transition duration; #278 Removed unnecessary Selectors.hasOwnProperty checks in animint.js"))
 Description: Functions are provided for defining animated,
  interactive data visualizations in R code, and rendering
  on a web page. The 2018 Journal of Computational and 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -64,7 +64,7 @@ Authors@R: c(
      comment="Animint2 GSoC 2025"),
     person("Gaurav", "Chaudhary",
      role="ctb",
-     comment="Remove unused css.file parameter; fix issue #233 selector values; fix issue #273 update_axes tick font-size; fix issue #276 update_axes transition duration"),
+     comment="Remove unused css.file parameter; fix issue #233 selector values; fix issue #273 update_axes tick font-size; fix issue #276 update_axes transition duration; #278 Removed unnecessary Selectors.hasOwnProperty checks in animint.js"),
     person("Nandani", "Aggarwal",
      role="ctb",
      comment="Removed geom_dotplot and replaced with error"))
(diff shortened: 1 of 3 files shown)


inst/htmljs/animint.js
```diff
@@ -1567,7 +1567,7 @@ var animint = function (to_select, json_file) {
           eActions = function(groups) {
             // Handle transitions seperately due to unique structure of geom_label_aligned
             var transitionDuration = 0;
-            if (Selectors.hasOwnProperty(selector_name)) {
+            if (selector_name) {
```
tdhock: remove if?

inst/htmljs/animint.js
```diff
@@ -1863,7 +1863,7 @@ var animint = function (to_select, json_file) {
           positionTooltip(tooltip, tooltip.html());
         });
     }
-    if(Selectors.hasOwnProperty(selector_name)){
+    if(selector_name){
```
tdhock: remove?

inst/htmljs/animint.js
```diff
@@ -2011,7 +2011,7 @@ var animint = function (to_select, json_file) {
     // update existing axis
     var xyaxis_sel = element.select("#"+viz_id+"_"+p_name).select("."+axes+"axis_"+panel_i);
     var milliseconds = 0;
-    if(v_name && Selectors.hasOwnProperty(v_name) && Selectors[v_name].hasOwnProperty("duration")){
+    if(v_name && Selectors[v_name].hasOwnProperty("duration")){
```
tdhock: remove v_name && ?

tdhock (changes requested):
also please investigate CI failure

tdhock (changes requested):
please revise

DESCRIPTION
```diff
@@ -64,7 +64,7 @@ Authors@R: c(
      comment="Animint2 GSoC 2025"),
     person("Gaurav", "Chaudhary",
      role="ctb",
-     comment="Remove unused css.file parameter; fix issue #233 selector values; fix issue #273 update_axes tick font-size; fix issue #276 update_axes transition duration"),
+     comment="Remove unused css.file parameter; fix issue #233 selector values; fix issue #273 update_axes tick font-size; fix issue #276 update_axes transition duration; #278 Removed unnecessary Selectors.hasOwnProperty checks in animint.js"),
```
tdhock:
```suggestion
     comment="GSOC 2026"),
```
please move PR attribution to NEWS?

inst/htmljs/animint.js
```diff
@@ -101,6 +101,12 @@ var animint = function (to_select, json_file) {
     return safe_name(selector_name) + "_variable";
   }
 
+  function selector_has_duration(name) {
+    return !!(name &&
```
tdhock: what is double bang (!!) ?

tdhock: why doesn’t `Selectors[name].hasOwnProperty("duration")` work?
we should never be doing `Selectors[unrecognized_name]`

ANAMASGARD:
Sir @tdhock  thanks for the review , 
 - !! removed — the function is only ever called inside if(...), so JavaScript already handles truthy/falsy correctly there. No need to force a boolean.

- Selectors.hasOwnProperty(name) removed — agree with your point. If name is a real selector, it should already be in Selectors, so checking membership here is redundant.

 - name && kept — this one is actually necessary. During initial rendering, draw_geom and geom_label_aligned run for every geom, including ones that have no selector at all. For those, selector_name is null. If I remove the name && guard, Selectors[null] throws a TypeError and crashes all JS tests. I confirmed this in an earlier commit where removing it caused all 149 JS tests to fail.

tdhock (approved):

PR #315, opened 2026-03-26
 test-renderer-bisection-method added

Nishita-shah1:
## hard test solution for GSoC 2026.

Adds unit tests covering both static rendering and interactive browser behaviour for a Bisection Method animation built with `animint2` and tested in a live Chrome session via Chromote.

Video Link: https://youtu.be/CWurjqqPWLA

Mentors @tdhock , @suhaani-agarwal and @Faye-yufan could you please review this test and give your feedback.
Thank You so much!

### DIFF
--- a/tests/testthat/test-renderer-bisection-method.R
+++ b/tests/testthat/test-renderer-bisection-method.R
@@ -1,0 +1,179 @@
+library(testthat)
+library(animint2)
+library(XML)
+
+context("Bisection Method Animation renderer")
+
+tests_init()
+
+# ----------- DATA SETUP -----------
+
+set.seed(42)
+max_iterations <- 10
+
+# f(x) = x^3 - x - 2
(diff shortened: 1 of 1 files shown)


suhaani-agarwal:
the interactive tests look good, but in the video it would be better to show a step-by-step execution instead of running everything together , specially where you have used clickID() . for example, you could demonstrate verifying the state before a click and then after a specific click, showing the change in the visualization when using clickID().

Nishita-shah1:
Okay, I'll edit the video with new changes and update you.

Nishita-shah1:
@suhaani-agarwal https://www.youtube.com/watch?v=htay3fXV-7s I have explained each test in very detail in this video, as I couldn't edit the previous video after it being uploaded.
Thank you for the feedback and mentorship!

PR #316, opened 2026-03-29
Add renderer test for simple random sampling

hargun144:
This PR adds a renderer test for a Simple Random Sampling visualization 
based on the animation package's sample.simple() function.

The test verifies that animint2 correctly renders at least 100 circle 
elements (the 10x10 population grid) in the SVG when the viz is loaded.

Related to my GSoC 2026 application tests:
- Live viz: https://hargun144.github.io/animint2-medium
- Source: https://github.com/hargun144/animint2-test

### DIFF
--- a/tests/testthat/test-renderer-simple-random-sampling.R
+++ b/tests/testthat/test-renderer-simple-random-sampling.R
@@ -1,0 +1,51 @@
+library(animint2)
+library(XML)
+
+set.seed(42)
+nrow_val <- 10
+ncol_val <- 10
+size <- 15
+nmax <- 20
+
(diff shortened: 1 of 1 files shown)


suhaani-agarwal:
Hi @hargun144 , the renderer test code should show interaction using clickID() ,It would be helpful if the test shows a step-by-step execution , for example, verifying the state of browser viz before a click and then after it , so that the interaction behavior is clearly tested, not just the initial render.

hargun144:
Updated the test to verify state before and after clickID() interaction as suggested — now passes with 2 successes. Please take a look

PR #319, opened 2026-03-30
test: add geom_bar regression test for update_axes limitation

AZ9tumas:
## update_axes test for geom_bar
Through this PR I aim to add a test documenting the current `update_axes` issue for `geom_bar`.

While going through pr #48, I traced the current compiler-side domain computation and found that `compute_domains()` behaves differently for `geom_bar` and `geom_point` since they require different columns to compute domains.

From my understanding, in this code section
https://github.com/animint/animint2/blob/543d0d9475b9fc1a34fae5899928251b1bc3ddc3/R/z_animintHelpers.R#L403-L414

for `geom_point` it uses `x` and `y` but for `geom_bar` it uses `xmin`, `xmax`, `ymin`, `ymax`. 
The former is present in `built_data` but not the latter. This is connected to this code section:
https://github.com/animint/animint2/blob/543d0d9475b9fc1a34fae5899928251b1bc3ddc3/R/z_animint.R#L511-L515

Since the column names are not found, the `else if` block is executed and `compute_domains` returns `NULL`. I have also added a warning for the same:
```r
  }else if(!all(use_cols %in% names(built_data))){
    missing <- use_cols[!use_cols %in% names(built_data)]
    warning(paste0(
    "Cannot compute axis domain for geom_", geom_name,
    ": column(s) ", paste(missing, collapse=", "),
    " not found in the pre-stat layer data."), call. = FALSE)
    return(NULL)
  }
```

## My implementation approach
- OO approach as suggested and continue working on the Geom class refactor from #107. 
- Add geom-level domain computation in the renderer from the available data (.tsv)
- Handle geom-specific logic cleanly in the corresponding geom class rather than in compiler-side helper code.
- Modify `update_scales()` so it uses renderer-computed domains instead of relying only on compiler-saved `axis_domains`
- Eventually generalize to other geoms that currently use different domain columns.
- More tests that check for axes updates after clicks throughout other geoms.

### DIFF
--- a/R/z_animintHelpers.R
+++ b/R/z_animintHelpers.R
@@ -418,6 +418,11 @@ compute_domains <- function(built_data, axes, geom_name,
                    geom_name), call. = FALSE)
     return(NULL)
   }else if(!all(use_cols %in% names(built_data))){
+    missing <- use_cols[!use_cols %in% names(built_data)]
+    warning(paste0(
+    "Cannot compute axis domain for geom_", geom_name,
+    ": column(s) ", paste(missing, collapse=", "),
+    " not found in the pre-stat layer data."), call. = FALSE)
     return(NULL)
   }
   domain_vals <- list()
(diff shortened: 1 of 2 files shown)


AZ9tumas:
@tdhock @suhaani-agarwal Could you please review this and suggest? 😁

PR #320, opened 2026-04-04
fix issue #252, support subgroup aesthetic in geom_polygon for holes

### LINKED ISSUE #252: draw polygon / path with holes
this is apparently possible with d3 https://devdocs.io/d3~3/geo-paths.md#_path

and it can be done with R grid / isoband
```r
# shape with hole
m <- matrix(c(0, 0, 0, 0, 0, 0,
              0, 1, 1, 1, 1, 0,
              0, 1, 0, 0, 1, 0,
              0, 1, 0, 0, 1, 0,
              0, 1, 1, 1, 1, 0,
              0, 0, 0, 0, 0, 0), 6, 6, byrow = TRUE)
isoband::plot_iso(m, 0.5, 1.5)
```

Image

this animint code gives two filled polygons instead of one with a hole.
```r
res=isoband::isobands((1:ncol(m))/(ncol(m)+1), (nrow(m):1)/(nrow(m)+1), m, 0.5, 1.5)[[1]]
## ?grid.path
##     rule: A character value specifying the fill rule: either
##           ‘"winding"’ or ‘"evenodd"’.
## A path is like a polygon except that the former can contain holes,
## as interpreted by the fill rule; these fill a region if the path
## border encircles it an odd or non-zero number of times,
## respectively.
viz <- animint(poly=ggplot()+
  geom_polygon(aes(
    x, y, group=id, fill=ID),
    data=as.data.table(res)[, ID := factor(id)]))
viz$poly
viz

> res
$x
 [1] 0.7142857 0.5714286 0.4285714 0.2857143 0.2142857 0.2142857 0.2142857 0.2142857 0.2857143
[10] 0.4285714 0.5714286 0.7142857 0.7857143 0.7857143 0.7857143 0.7857143 0.5714286 0.6428571
[19] 0.6428571 0.5714286 0.4285714 0.3571429 0.3571429 0.4285714

$y
 [1] 0.2142857 0.2142857 0.2142857 0.2142857 0.2857143 0.4285714 0.5714286 0.7142857 0.7857143
[10] 0.7857143 0.7857143 0.7857143 0.7142857 0.5714286 0.4285714 0.2857143 0.3571429 0.4285714
[19] 0.5714286 0.6428571 0.6428571 0.5714286 0.4285714 0.3571429

$id
 [1] 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 2 2 2 2 2 2 2 2
```

Image

Image

Nishita-shah1:
Fixes #252

## Problem
animint2 rendered isoband hole polygons as two separate filled shapes
instead of one polygon with a transparent hole. The `subgroup` aesthetic
(added to ggplot2 in 2019) was completely unknown to animint2's compiler
and renderer.

Real-world use case: Hi-C genomic cluster contour visualization where
holes in contour polygons are scientifically meaningful.

## After (fix)
Screenshot 2026-04-04 200117

Screenshot 2026-04-04 200133

## Solution
Three-layer fix across the full pipeline:

**R/geom-polygon.r** - added `subgroup = NULL` to `default_aes` so
GeomPolygon accepts the aesthetic.

**R/geom-.r** -- when `subgroup` is present in polygon data:
- sets `g$data_has_subgroup <- TRUE` (serialized into plot.json)
- converts subgroup to character so it survives TSV write
- adds it to `g$types` so JS type converter keeps it as string

**inst/htmljs/animint.js** - when `data_has_subgroup` is true, uses
`d3.geo.path().projection(null)` with GeoJSON Polygon format instead
of `d3.svg.line()`. GeoJSON coordinates map directly to subgroup
structure: `[outerRing, holeRing1, ...]`. SVG `fill-rule: evenodd`
ensures hole rings appear transparent.

## Tests
17 tests, 0 failures covering:
- compiler: subgroup in TSV, data_has_subgroup in plot.json,
  no flag without subgroup
- renderer: SVG path used, multiple M commands in d attribute,
  evenodd fill-rule applied
- interactive: clickID and path checks for all 3 polygon cases
  (hole_and_mid, only_hole, no_hole) as specified in the issue
Screenshot 2026-04-04 201101

@tdhock @suhaani-agarwal @Faye-yufan  could you please review this, thank you! Fixes #252


Nishita-shah1:
Hi @tdhock , @Faye-yufan and @suhaani-agarwal . Could you please review this. 
Thank You!

inst/htmljs/animint.js
```diff
@@ -1371,11 +1371,44 @@ var animint = function (to_select, json_file) {
...
+              if(ring.length > 0){
+                ring = ring.concat([ring[0]]);
+              }
+              return ring;
+            });
+            var geojson = {
+              type: "Polygon",
+              coordinates: coords
+            };
+            return geoPath(geojson);
```
tdhock: I do not understand this code.
I was expecting some library function call here, instead of these forEach/map calls. is that possible?
or can you please add comments and/or link some documentation?

tdhock (approved):

inst/htmljs/animint.js
Nishita-shah1: forEach and map here are plain JavaScript built-in array methods, forEach loops over each point to group them by subgroup into rings_map, and map transforms those groups into closed GeoJSON rings.

Is there any other library function to use? Could you please tell me about it? I'll read about it and make the changes in the code.

Also, I looked into d3.group() (https://d3js.org/d3-array/group) as a possible replacement for the forEach grouping step. Would that be a good direction to explore?

inst/htmljs/animint.js
```diff
@@ -1371,11 +1371,44 @@ var animint = function (to_select, json_file) {
...
+      // polygon with subgroup aesthetic: use d3.geo.path with evenodd
+      if(g_info.geom === "polygon" && g_info.data_has_subgroup){
+        var geoPath = d3.geo.path().projection(null);
+        eActions = function(e){
+          e.attr("d", function(d){
+            var points = keyed_data[d.value];
+            var rings_map = {};
+            var ring_order = [];
+            points.forEach(function(pt){
+              var sg = pt.subgroup !== undefined ? pt.subgroup : "1";
```
Faye-yufan: Hi @Nishita-shah1 , would g.data pass in null/na if subgroup is true? In what scenario that subgroup == undefined but g_info.data_has_subgroup == true?
Any test case for this scenario?

inst/htmljs/animint.js
suhaani-agarwal: > Also, I looked into d3.group() (https://d3js.org/d3-array/group) as a possible replacement for the forEach grouping step. Would that be a good direction to explore?

hi nishita! d3.group() is from D3v7+, but animint.js uses D3v3. take a look at how grouping is done elsewhere in this file (search for d3.nest). The same pattern could likely apply to your code.

you can try refactoring the forEach grouping block using the existing library pattern you find.

inst/htmljs/animint.js
Nishita-shah1: Thank you mentors for the feedback .
I have updated the code and now it uses d3.nest.  
https://docs.google.com/document/d/1QBB-OU3Gjli9TBWmX5k7ER3R3JqrCDDGI1usf22027I/edit?usp=sharing
in this doc i have written my logic and all the test cases.

Nishita-shah1: > Hi @Nishita-shah1 , would g.data pass in null/na if subgroup is true? In what scenario that subgroup == undefined but g_info.data_has_subgroup == true? Any test case for this scenario?

Hi @Faye-yufan , i think it cannot happen because :
Step 1- 
(R - geom-.r):data_has_subgroup is set ONLY when 'subgroup' %in% names(g.data) → column must physically exist in the data frame
Step 2- (JS — animint.js):
if(g_info.geom==='polygon' && g_info.data_has_subgroup), nly runs when R confirmed the column exists, pt.subgroup will always be a string ('1', '2', 'NA', etc.), never JS undefined

I have explained my logic and all the test cases in this doc https://docs.google.com/document/d/1QBB-OU3Gjli9TBWmX5k7ER3R3JqrCDDGI1usf22027I/edit?tab=t.0

Nishita-shah1:
Hi @tdhock , @suhaani-agarwal and @Faye-yufan ,
Update on PR #320 - I have made the following changes based on the review feedback:
- Replaced `forEach/rings_map` with `d3.nest()` as suggested by Suhaani (D3v3 compatible)
- Removed the `pt.subgroup` undefined fallback as flagged by Yufan
 - Added documentation links with my logic approach and test cases
 
 Regarding the CI failures - the original PR #320 was raised via my fork (`Nishita-shah1/animint2`), due to which GitHub blocks secrets from fork PRs and checks were failing. To fix this, I raised a new PR directly from the `animint/animint2` branch:
 https://github.com/animint/animint2/pull/326
 In PR #326, CRAN and R_coverage are now passing 
 The only remaining failing check is **JS_coverage**.

 After checking the logs, I believe the issue is that my renderer tests use chromote (headless Chrome) to test the SVG output, and the CI Linux server has no virtual display:
> ```
> Warning: Failed to save JS coverage: code: -32000
> Missing X server or $DISPLAY
> The platform failed to initialize. Exiting.
> ```
>
 Could you please advise on the best way to resolve this? I see two possible approaches:
 1. Add `xvfb-run -a` to the `tests.yaml` JS_coverage run step so Chrome has a virtual display
 2. Add `skip_if_not(has_display())` to the renderer/interactive tests so they are skipped gracefully on CI when no display is available

 Happy to implement whichever approach you prefer. Thank you so much for your continued guidance!

tdhock:
closing in favor of #326

PR #326, opened 2026-05-17
Fix/252 polygon holes subgroup

### LINKED ISSUE #252: draw polygon / path with holes

Nishita-shah1:
Closes #252 

creating pr from branch instead of from fork


tests/testthat/test-renderer-polygon-holes.R
```diff
@@ -0,0 +1,229 @@
...
+  ## should not add or remove polygon path elements
+  expect_equal(count.before, count.after)
+})
+
+test_that("interactive: all rendered path elements have non-empty d attribute", {
+  html      <- getHTML()
+  path.list <- getNodeSet(html, '//g[contains(@class,"geom")]//path')
+  expect_true(length(path.list) >= 1)
+  d.vals <- sapply(path.list, function(node) xmlGetAttr(node, "d"))
+  expect_true(
```
tdhock, line 214: can you please avoid expect_true, and use expect_ something more specific like equal?

tdhock (changes requested):
can you also please post screenshots of the test cases?

Nishita-shah1:
> can you also please post screenshots of the test cases?
Hi @tdhock ,
these are the screenshots of the test-renderer-polygon-holes.R file test cases.
image
image

tdhock:
I tried this small-scale real data application, https://github.com/tdhock/hicream-viz/blob/main/data-2025-10-09/figure-pixels-chr1-Cluster73-fix.R and this PR works!

image

top: this PR.
bottom: master branch.

tdhock:
here is a rendered animint using this PR https://tdhock.github.io/2026-05-22-HiC-pixels-chr1-Cluster73-subgroup/
image

and the same animint which does not render correctly using current master https://tdhock.github.io/2026-05-22-HiC-pixels-chr1-Cluster73-bad/
image

this is evidence of a successful implementation, great work, thanks very much Nishita!

tdhock:
currently there is only one clickID in tests. are these tests strong enough to fail on current master?

the issue #252 had a screenshot of master drawing the test case like this:
image
we should add a clickID or mouseMove to the test so that it fails when drawn with current master.
for example, if we put clickSelects and showSelected on the polygon, and draw a red dot behind the polygon in the area that it is incorrectly drawn (hold_and_mid_mid_bottom), then clickID on the red dot, it will actually click on the polygon, which will hide the polygon, and so we should add an expectation that the polygon is still visible even after that click. Does that make sense?

tests/testthat/test-renderer-polygon-holes.R
```diff
@@ -0,0 +1,229 @@
+library(testthat)
+library(animint2)
+library(XML)
+context("Polygon holes via subgroup aesthetic")
+tests_init()
```
tdhock, line 5: please delete all this from test files

tests/testthat/test-renderer-polygon-holes.R
```diff
@@ -0,0 +1,229 @@
...
+    (nrow(m):1)/(nrow(m)+1),
+    m, 0.5, 1.5)[[1]]
+  as.data.frame(res)
+}
+
+## full test case from issue: 3 polygon types side by side
+## hole_and_mid: outer ring + hole + island (3 subgroups)
+## only_hole:    outer ring + hole (2 subgroups)
+## no_hole:      outer ring only (1 subgroup)
+make.full.data <- function(){
```
tdhock, line 29: it is not necessary (and potentially confusing) to create a function if you only call it once.

tests/testthat/test-renderer-polygon-holes.R
```diff
@@ -0,0 +1,229 @@
...
+## simple donut: outer ring (id=1) + hole (id=2)
+make.hole.data <- function(){
+  m <- matrix(c(
+    0,0,0,0,0,0,
+    0,1,1,1,1,0,
+    0,1,0,0,1,0,
+    0,1,0,0,1,0,
+    0,1,1,1,1,0,
+    0,0,0,0,0,0), 6, 6, byrow=TRUE)
+  res <- isoband::isobands(
```
tdhock, line 18: we should use contourLines() instead of isobands() so we can avoid adding a dependency.

tests/testthat/test-renderer-polygon-holes.R
Nishita-shah1: Sure, I'll update it.

Nishita-shah1:
> currently there is only one clickID in tests. are these tests strong enough to fail on current master?
> 
> the issue #252 had a screenshot of master drawing the test case like this: image we should add a clickID or mouseMove to the test so that it fails when drawn with current master. for example, if we put clickSelects and showSelected on the polygon, and draw a red dot behind the polygon in the area that it is incorrectly drawn (hold_and_mid_mid_bottom), then clickID on the red dot, it will actually click on the polygon, which will hide the polygon, and so we should add an expectation that the polygon is still visible even after that click. Does that make sense?

yes, we can do that. I'll add some more tests.

Nishita-shah1:
> here is a rendered animint using this PR https://tdhock.github.io/2026-05-22-HiC-pixels-chr1-Cluster73-subgroup/ image
> 
> and the same animint which does not render correctly using current master https://tdhock.github.io/2026-05-22-HiC-pixels-chr1-Cluster73-bad/ image
> 
> this is evidence of a successful implementation, great work, thanks very much Nishita!

Thank you @tdhock , I surely update the tests based on the feedbacks and correct the naming of the tests.

tdhock:
actually can you please just close this PR and port your new JS code to #328 ?

tdhock:
please do keep those review comments in mind for future PRs though.

Nishita-shah1:
Sure @tdhock , I'll close this PR and port to #328 . 
Yes, I'll keep in mind the review comments for future PRs.

Issue #325, opened 2026-05-13
need to copy animint dir to _site after quarto render


tdhock:
rendering animints in R code chunks in quarto sites only works if we copy animint directories from the source folder to the output _site folder, as in
https://github.com/animint/animint-blog/blob/add-netlify/render.R
https://github.com/animint/animint-manual-en/blob/main/render.R

It is not great to repeat this logic in every project’s render.R script.
It would be preferable if we could move this code into the animint2 package, and quarto could somehow "automatically" know to copy these files to _site.

tdhock:
probably the solution involves `knit_print.animint`, https://github.com/animint/animint2/blob/master/R/z_knitr.R

tdhock:
documentation is here https://cran.r-project.org/web/packages/knitr/vignettes/knit_print.html
but it looks like we need some quarto option to support projects with multiple qmd files (not just knitr which seems to support only one source file at a time).

ashishtiwari03:
I opened PR #331 with a failing test for this issue.

  The test renders a minimal Quarto website with one animint plot and checks whether `_site/myplot/plot.json` exists
  after `quarto::quarto_render()`.

  I have not pushed the fix yet, so the PR currently only contains the failing test.

tdhock:
great thanks

PR #331, opened 2026-05-24
Add failing test for quarto animint output copy

ashishtiwari03:
### Description

  This PR adds a failing test for #325.

  The test creates a small Quarto website project with one animint plot, renders it with `quarto::quarto_render()`, and checks
  for the generated animint files in `_site/myplot/`.

  On current `master`, the animint directory is created during knitting but is not copied into the Quarto website output
  directory, so the rendered site is missing files such as `plot.json`.

  No fix or NEWS entry is included yet. This is only the failing test commit, following the test-first workflow described in
  the contributing guide.

### DIFF
--- a/R/z_knitr.R
+++ b/R/z_knitr.R
@@ -12,6 +12,7 @@ knit_print.animint <- function(x, options, ...) {
   viz_id <- gsub("[^[:alnum:]]", "", options$label)
   out.dir <- file.path(output.dir, viz_id)
   animint2dir(x, out.dir = out.dir, open.browser = FALSE)
+  copy_quarto_animint_dir(out.dir, viz_id)
   res <- if(knitr::is_latex_output())sprintf(
     "\\IfFileExists{%s/Capture.PNG}{\\includegraphics[width=\\textwidth]{%s/Capture.PNG}}{}", out.dir, out.dir
   ) else sprintf(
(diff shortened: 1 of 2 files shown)


R/z_knitr.R
```diff
@@ -36,6 +37,33 @@ knit_print.animint <- function(x, options, ...) {
   knitr::asis_output(res, meta = list(animint = structure("", class = "animint")))
 }
 
+quarto_output_dir <- function(project.root) {
+  quarto.yml <- file.path(project.root, "_quarto.yml")
+  if (!file.exists(quarto.yml)) return(NULL)
+
+  yml.lines <- readLines(quarto.yml, warn = FALSE)
+  output.line <- grep("^[[:space:]]+output-dir:", yml.lines, value = TRUE)
```
tdhock: this seems unusual. can you please post an issue on quarto to ask how we should do this?

ashishtiwari03: Thanks, that makes sense. I opened an issue on Quarto to ask for the recommended way to get the project output
  directory from inside `knit_print()`:  
https://github.com/orgs/quarto-dev/discussions/14545

ashishtiwari03: Quarto maintainers replied here:

  https://github.com/orgs/quarto-dev/discussions/14545

  They suggested that `fig.path` is probably not the right mechanism, and that dependencies associated with output
  should probably go through `knit_meta` / `knit_print()` metadata, possibly via `htmltools` dependencies.

  So I agree the current `_quarto.yml` parsing in this PR is probably not the right final approach. I can try revising
  the fix in that direction.

tdhock: ok thanks.
please try to do it without htmltools.

ashishtiwari03: Based on the Quarto discussion, I will try revising this without `htmltools` and without parsing `_quarto.yml`.

  I will look at using `knitr::asis_output(..., meta = ...)` / `knit_meta` directly, since `knit_print.animint()`
  already returns metadata.

ashishtiwari03: cderv added another note in the Quarto discussion:

  https://github.com/orgs/quarto-dev/discussions/14545

They suggested looking at how other packages ship HTML dependencies with knitr output. They mentioned that in rmarkdown contexts this is usually done through htmlwidgets or html dependencies.

Since you asked to try without `htmltools`, I will first check whether this can be done using `knit_meta` directly,
 without adding htmltools-based dependencies.

ashishtiwari03: I revised the fix to avoid parsing _quarto.yml and to avoid htmltools.

The updated knit_print.animint() now leaves Quarto/rmarkdown to discover the generated animint files through plain HTML resource links emitted with the chunk output. The existing knitr metadata flag is still used only to avoid injecting the shared JS/CSS more than once.

  I also added an assertion that the rendered Quarto site contains myplot/animint.js in addition to myplot/plot.json.

  Locally, the focused Quarto test passes:
  FAIL 0 | WARN 0 | SKIP 0 | PASS 2

PR #292, opened 2026-01-23
Fix #140: Add ability to disable automatic showSelected on legends

ANAMASGARD:
## Issue
FIXES #140

## Problem
Using `aes(color=variable)` forces `showSelected` behavior on the legend with no opt-out.

## This PR { First commit }
Adds failing test proposing `showSelected=character()` to disable automatic selector.

Screenshot From 2026-01-23 18-03-07

PR #336, opened 2026-06-05
Fix #333: allow showSelected.legend=FALSE to keep explicit showSelected without legend auto-injection

### LINKED ISSUE #333: opt out of auto but specify other showSelected
#292 works for the case where you want no showSelected at all.
But what if we want to opt out of the showSelected legend, but we want to specify another variable in showSelected?
@ANAMASGARD can you please investigate?

* create a minimal test that fails for current code
* what API could be used??

ANAMASGARD:
## Summary

Fixes https://github.com/animint/animint2/issues/333

PR #292 added `showSelected = character()` so a layer can opt out of **all** auto legend `showSelected` behavior while keeping the legend visible. Issue #333 asks for a narrower case: keep an **explicit** `showSelected` (e.g. filter by `year`) but **do not** auto-add `showSelected` for colour/fill legend variables on the same layer (e.g. `region` from `colour = region`).

## What was wrong?

On master, if you write `geom_point(..., colour = region, showSelected = "year")`, animint2 adds **both** `year` and `region` to the layer’s `showSelected` aesthetics. There was no way to suppress only the legend-driven part while keeping your own `showSelected`.

## What this PR does

**Commit 1 (this push): failing test only** — per [CONTRIBUTING](https://github.com/animint/animint2/blob/master/.github/CONTRIBUTING.md)

- Adds `tests/testthat/test-renderer2-issue333-opt-out-ss-with-other-showSelected.R`
- Uses proposed API: `geom(..., showSelected = "year", showSelected.legend = FALSE)`
- Asserts (once the API works):
  - Viz compiles
  - Layer `showSelected` is only `"year"` (not `"region"`)
  - Colour legend for `region` is still present
  - `year` selector stays `single` for animation / tallrect
- **Expected CI on commit 1:** red — error at `animint2dir()` because `error_for_showSelected_variants()` does not whitelist `showSelected.legend` yet

**Commit 2 (next): implementation**

- Whitelist `showSelected.legend` in `error_for_showSelected_variants()` (`R/z_animintHelpers.R`)
- In `addShowSelectedForLegend()`, when `isFALSE(extra_params$showSelected.legend)`, skip appending legend variables to `showSelected` (still honor user-specified `showSelected`)
- Strip `showSelected.legend` in `getLayerParams()` so it never appears in TSV/JSON export
- **Expected CI after commit 2:** green — same test file passes

## Expected result

Users can write:

```r
geom_point(aes(..., colour = region), showSelected = "year", showSelected.legend = FALSE)

### DIFF
--- a/R/z_animintHelpers.R
+++ b/R/z_animintHelpers.R
@@ -6,11 +6,7 @@
 #' @param L layer of the plot
 #' @return L : Layer with additional mapping to new aesthetic
 addShowSelectedForLegend <- function(meta, legend, L){
-  ## Check if user explicitly disabled showSelected with character()
-  ## If showSelected is character(0), user wants to opt out of auto-showSelected
-  user_disabled_showSelected <- is.character(L$extra_params$showSelected) &&
-    length(L$extra_params$showSelected) == 0
-  user_disabled_legend_showSelected <- isFALSE(L$extra_params$showSelected.legend)
+  legend_auto_ss_disabled <- isFALSE(L$extra_params$showSelected.legend)
   for(legend.i in seq_along(legend)) {
     one.legend <- legend[[legend.i]]
     ## the name of the selection variable used in this legend.
@@ -35,9 +31,7 @@ addShowSelectedForLegend <- function(meta, legend, L){
         ## only add showSelected aesthetic if the variable is
         ## used by the geom
         type.vec <- one.legend$legend_type
-        if((!user_disabled_showSelected) &&
-           (!user_disabled_legend_showSelected) &&
-           any(type.vec %in% names(L$mapping))){
+        if(!legend_auto_ss_disabled && any(type.vec %in% names(L$mapping))){
           L$extra_params$showSelected <- c(L$extra_params$showSelected, s.name)
         }
       }
--- a/R/z_animintHelpers.R
+++ b/R/z_animintHelpers.R
@@ -10,6 +10,7 @@ addShowSelectedForLegend <- function(meta, legend, L){
   ## If showSelected is character(0), user wants to opt out of auto-showSelected
   user_disabled_showSelected <- is.character(L$extra_params$showSelected) &&
     length(L$extra_params$showSelected) == 0
+  user_disabled_legend_showSelected <- isFALSE(L$extra_params$showSelected.legend)
   for(legend.i in seq_along(legend)) {
     one.legend <- legend[[legend.i]]
     ## the name of the selection variable used in this legend.
@@ -34,7 +35,9 @@ addShowSelectedForLegend <- function(meta, legend, L){
         ## only add showSelected aesthetic if the variable is
         ## used by the geom
         type.vec <- one.legend$legend_type
-        if((!user_disabled_showSelected) && any(type.vec %in% names(L$mapping))){
+        if((!user_disabled_showSelected) &&
+           (!user_disabled_legend_showSelected) &&
+           any(type.vec %in% names(L$mapping))){
           L$extra_params$showSelected <- c(L$extra_params$showSelected, s.name)
         }
       }
@@ -189,7 +192,10 @@ hjust2anchor <- function(hjust){
 error_for_showSelected_variants <- function(params) {
   if (is.null(names(params))) return(NULL)
   # Match any parameter that starts with "showSelected" but is not exactly "showSelected"
-  invalid_showSelected <- grep("^showSelected.+", names(params), value = TRUE)
+  invalid_showSelected <- setdiff(
+    grep("^showSelected.+", names(params), value = TRUE),
+    "showSelected.legend"
+  )
   if (length(invalid_showSelected) > 0) {
     stop(sprintf("Invalid parameter(s): %s. Please use geom(showSelected = character_vector_of_variable_names)",
         paste(invalid_showSelected, collapse = ", ")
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by ANAMASGARD** [comment]
Sir  @tdhock, I’ve implemented the fix for issue :
- Added showSelected.legend = FALSE to opt out of legend-driven showSelected auto-injection.
- Kept explicit showSelected = "year" working as intended.
- Preserved the existing showSelected = character() opt-out behavior from PR #292.
- Whitelisted showSelected.legend in parameter validation.
- Removed showSelected.legend before export so it stays compiler-only.
- Added a regression test for the new behavior.
Screenshot From 2026-06-07 22-17-32

**Comment by tdhock** [comment]
it seems reasonable to create a new argument, `showSelected.legend`, for this.
in that case, I would expect `showSelected=character()` no longer has special meaning, is that right?

**Comment by ANAMASGARD** [comment]
Yes, that is my understanding as well. I’ll consolidate on showSelected.legend=FALSE as the explicit opt-out, remove the special handling for showSelected=character(), migrate the recent tests to the new API, and note the API cleanup in NEWS.

File: R/z_animintHelpers.R [hunk 1]
```diff
@@ -189,7 +186,10 @@ hjust2anchor <- function(hjust){
 error_for_showSelected_variants <- function(params) {
   if (is.null(names(params))) return(NULL)
   # Match any parameter that starts with "showSelected" but is not exactly "showSelected"
-  invalid_showSelected <- grep("^showSelected.+", names(params), value = TRUE)
+  invalid_showSelected <- setdiff(
+    grep("^showSelected.+", names(params), value = TRUE),
+    "showSelected.legend"
```
tdhock: I don’t think this is necessary. Can you please revert?

tests/testthat/test-renderer2-legend-without-showSelected.R
```diff
@@ -50,3 +48,16 @@ test_that("showSelected=character() keeps legend selector but opts layer out", {
   show_aes_names <- names(info$geoms[[point_geom_name]]$aes)
   expect_identical(any(grepl("^showSelected", show_aes_names)), FALSE)
 })
+test_that("showSelected=character() no longer opts layer out of legend injection", {
+  viz_char_ss <- list(
+    plot1 = ggplot(test_data, aes(x, y, color = comparison)) +
```
tdhock: use informative name instead of number (plot1)

tdhock (changes requested):

R/z_animintHelpers.R
ANAMASGARD: Done

R/z_animintHelpers.R
```diff
@@ -6,10 +6,7 @@
 #' @param L layer of the plot
 #' @return L : Layer with additional mapping to new aesthetic
 addShowSelectedForLegend <- function(meta, legend, L){
-  ## Check if user explicitly disabled showSelected with character()
-  ## If showSelected is character(0), user wants to opt out of auto-showSelected
-  user_disabled_showSelected <- is.character(L$extra_params$showSelected) &&
-    length(L$extra_params$showSelected) == 0
+  legend_auto_ss_disabled <- isFALSE(L$extra_params$showSelected.legend)
```
Faye-yufan, line 9: Hi @ANAMASGARD , Can you create a test case that `showSelected.legend=TRUE`?

ANAMASGARD: Hi @Faye-yufan thanks for review 

I have added  test for showSelected.legend = TRUE in test-renderer2-legend-without-showSelected.R. It checks that legend-driven showSelected still works like the default.

Also fixed a small bug: extra_params$showSelected was accidentally picking up showSelected.legend when TRUE was passed. Using extra_params[["showSelected"]] fixed it.

tdhock:
R tests fail?

ANAMASGARD:
@tdhock It was failing but after I re ran the CI it passed this time .

tests/testthat/test-renderer2-issue333-opt-out-ss-with-other-showSelected.R
```diff
@@ -0,0 +1,46 @@
...
+info <- animint2dir(issue333_viz(), open.browser = FALSE)
+test_that("issue333 viz compiles with showSelected.legend=FALSE", {
+  expect_identical(info$time$variable, "year")
+})
+test_that("explicit showSelected=year without legend-injected region", {
+  expect_identical(showSelected_variables(info), "year")
+})
+test_that("colour legend for region is still compiled", {
+  legend.info <- info$plots$scatter$legend$region
+  expect_false(is.null(legend.info))
```
tdhock: un-necessary, reduntdant with below, please remove

tests/testthat/test-renderer2-legend-without-showSelected.R
```diff
@@ -49,4 +47,35 @@ test_that("showSelected=character() keeps legend selector but opts layer out", {
...
+  )
+  info <- animint2dir(viz, open.browser = FALSE)
+  legend_info <- info$plots$plot1$legend$comparison
+  expect_gt(length(legend_info$entries), 0)
+  expect_identical(legend_info$selector, "comparison")
+  geom_names <- names(info$geoms)
+  point_geom_name <- geom_names[grepl("_point_plot1$", geom_names)]
+  expect_identical(length(point_geom_name), 1L)
+  show_aes_names <- names(info$geoms[[point_geom_name]]$aes)
+  expect_identical(any(grepl("^showSelected", show_aes_names)), TRUE)
```
tdhock: expect_match

tdhock (changes requested):

tdhock (approved):
ok thanks

tdhock:
please resolve conflicts and merge?

ANAMASGARD:
Sir @tdhock  I have resolved the conflict please Megre the PR

ANAMASGARD:
@tdhock I had resolved the conflict so as you said I have squashed commits and merged the PR .

PR #311, opened 2026-03-08
Removed geom_dotplot and replace with error

NandaniAggarwal:
Fixes [#289](https://github.com/animint/animint2/issues/289)

Following @tdhock's suggestion, this PR removes `geom_dotplot()` 
and replaces it with a hard error directing users to `geom_point()`.

## Changes
- `geom_dotplot()` now calls `stop()` immediately
-  Added tests for error and geom_point alternative
- updated NEWS.md for this change

## Testing
All tests pass:

image

@tdhock please review and suggest

PR #341, opened 2026-06-13
fix: remove animint2 pre-install from atime workflow (issue #340)

### LINKED ISSUE #340: atime CI job fails on every PR due to broken \Sexpr in old baseline commit
The atime performance tests / comment job is failing on every PR. This is a CI infrastructure bug 

I encountered it in multiple of my PRs :-  
 - #253
- #261
- #285
- #336
## What happens
The atime job installs an old baseline commit (352f7e10040cb9de6ddd16416d342e9746c14c7a, version 2025.9.16) to compare performance before/after each PR. That old commit crashes during R CMD INSTALL with:
```
Error : geom_dotplot.Rd:125: processing build-stage \Sexpr code failed:
Error: No geom called GeomDotplot.
ERROR: installing Rd objects failed for package 'animint2'
Error in atime_versions_install(...) :
  '/opt/R/4.6.0/lib/R/bin/R' CMD INSTALL returned error status code 1
```
## Why ?
The old baseline commit's man/geom_dotplot.Rd has a \Sexpr at line 125 that looks up GeomDotplot at build time. GeomDotplot was removed from the package after that commit, so the install breaks. The current main branch already has a fixed geom_dotplot.Rd — this only breaks in the old baseline.

The PRs changes are never even reached - the job dies at baseline install time.

Image

ANAMASGARD:
Fixes #340

The atime `comment` job has been failing on every PR that touches `R/**` because the workflow pre-installed current `animint2` before the Autocomment action could install old baseline commits.

## Before / after
image

## Root cause

This is **not** a broken baseline commit. Commit `352f7e1` installs fine on its own.

The problem is step order in `atime.yaml`:

1. `R CMD INSTALL .` loads **current** `animint2` into `.libPaths` (no `GeomDotplot` since #311).
2. `atime_versions_install` later tries to install the Slow baseline as `animint2.352f7e1...`.
3. That old commit's `man/geom_dotplot.Rd` has `\Sexpr{animint2:::rd_aesthetics(...)}` at build time.
4. R resolves `animint2:::` to the **already-installed current package**, not the old tree being built.
5. Install fails: `Error: No geom called GeomDotplot`.

The Autocomment action already installs HEAD, merge-base, Slow, and Fast itself (each as `animint2.{sha}`). The removed step was not just redundant — it caused the collision.

## What changed

- Removed `R CMD INSTALL .` from `.github/workflows/atime.yaml`.
- Added a comment explaining why that step must not return (regression lock).
- Added `.github/workflows/atime.yaml` to the workflow `paths:` filter so this PR triggers and verifies the fix.
- **No change** to `Slow=` / `Fast=` in `.ci/atime/tests.R` — those SHAs are still correct for the #238 `getCommonChunk` comparison.

## Why no testthat file

As per contributing guidelines, we do not add tests whose primary purpose is CI infrastructure. The failing `comment` job **is** the test; this PR makes it pass. A `testthat` file that blacklists a SHA would not catch this mechanism if someone re-added the install step.

### DIFF
--- a/.ci/atime/tests.R
+++ b/.ci/atime/tests.R
@@ -1,3 +1,24 @@
+animint2_atime_pkg_edit <- function(old.Package, new.Package, sha, new.pkg.path) {
+  get("pkg.edit.default", envir=asNamespace("atime"))(
+    old.Package=old.Package,
+    new.Package=new.Package,
+    sha=sha,
+    new.pkg.path=new.pkg.path)
+  rd_replacement <- sprintf(
+    "get(\"rd_aesthetics\", envir=asNamespace(\"%s\"))",
+    new.Package)
+  rd_files <- Sys.glob(file.path(new.pkg.path, "man", "*.Rd"))
+  for (rd_file in rd_files) {
+    rd_lines <- readLines(rd_file)
+    new_lines <- gsub(
+      "animint2:::rd_aesthetics",
+      rd_replacement,
+      rd_lines,
+      fixed=TRUE)
+    if (!identical(rd_lines, new_lines)) writeLines(new_lines, rd_file)
+  }
+}
+
 test.list <- atime::atime_test_list(
   ## Adapted from https://github.com/animint/animint2/issues/235#issuecomment-3342083861
   "getCommonChunk improved in #238"=atime::atime_test(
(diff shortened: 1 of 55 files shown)


tdhock:
No obvious timing issues in HEAD=fix-atime-baseline-commit-340
[Comparison Plot](https://asset.cml.dev/97d9796ddc9ce3781f9a7b5955e25f08d9b5145b?cml=png\&cache-bypass=f31eb0e8-0e31-4b01-8be4-14e1c4df7304)

Generated via commit 6291631d0fa86f4d505138659ad6901b23de1541

Download link for the artifact containing the test results: [↓ atime-results.zip](https://github.com/animint/animint2/actions/runs/27460759874/artifacts/7609016478)

| Task | Duration |
|---|---|
| R setup and installing dependencies   | 2 minutes and 28 seconds       |
| Installing different package versions | 1 minutes and 0 seconds |
| Running and plotting the test cases   | 2 minutes and 8 seconds   |

tdhock:
I don’t think this is the right fix.
probably there should be a fix in atime, not animint2.

PR #344, opened 2026-08-18
Fix false Codecov project coverage drops by uploading R and JS reports together

### LINKED ISSUE #254: incorrect codecov failures
hi @suhaani-agarwal can you please help fix this issue?
If you look at recent commits to master https://github.com/animint/animint2/commits/master/
We see that I made a bunch of commits that only affect the documentation (CONTRIBUTING.md), not the code.
Yet the codecov report says that there was a decrease in project coverage, with respect to the previous push to master.
Image

* codecov/patch above is correct, there is no change to coverage in the patch.
* however codecov/project is obviously incorrect (doc changes do not affect code coverage), so there must be something wrong with the code coverage setup. Why does it say there is a decrease? how to fix?

does it have something to do with separate upload of R and JS coverage reports?

Nishita-shah1:
Fixes #254

## Summary
- Codecov `project` was failing on documentation-only commits because R and JS coverage were uploaded separately. If the JS job failed, an R-only report became the baseline, so the next full R+JS upload looked like a coverage drop.
- Coverage jobs now save artifacts instead of uploading immediately. A follow-up job uploads both reports together, and only if both R and JS jobs succeed.
- CRAN checks stay in their own job. Texlive is installed only there.

## Test plan
- [ ] Confirm CI runs `R_coverage` and `JS_coverage` in parallel, then `Upload combined coverage to Codecov`.
- [ ] Confirm the combined upload job is skipped if either coverage job fails.
- [ ] Confirm CRAN still runs independently.
- [ ] On the PR, Codecov `patch` and `project` should not fail just because one coverage job failed on a previous commit.

### DIFF
--- a/.github/scripts/save-r-coverage.R
+++ b/.github/scripts/save-r-coverage.R
@@ -1,0 +1,23 @@
+# Generate R coverage as Cobertura XML for a later combined Codecov upload.
+# Path cleanup uses gsub() so Codecov can match files in this repository.
+if (!requireNamespace("covr", quietly = TRUE)) {
+  stop("covr is required to generate R coverage")
+}
+
+cov <- covr::package_coverage(
+  quiet = FALSE,
+  type = "tests",
+  test_files = "tests/testthat.R"
+)
+print(cov)
+
+out_file <- "r-coverage.xml"
+covr::to_cobertura(cov, filename = out_file)
+
+root <- normalizePath(getwd(), winslash = "/", mustWork = TRUE)
+xml <- paste(readLines(out_file, warn = FALSE), collapse = "\n")
+xml <- gsub("\\\\", "/", xml)
+xml <- gsub(root, ".", xml, fixed = TRUE)
+writeLines(xml, out_file)
+
+message("Wrote ", out_file)
--- a/.github/scripts/save-r-coverage.R
+++ b/.github/scripts/save-r-coverage.R
@@ -1,5 +1,5 @@
-# Generate R coverage as Cobertura XML for a later combined Codecov upload.
-# Path cleanup uses gsub() so Codecov can match files in this repository.
+# Generate the same Rlang JSON that covr::codecov() used to POST, then save it
+# as an artifact for a later combined Codecov upload.
 if (!requireNamespace("covr", quietly = TRUE)) {
   stop("covr is required to generate R coverage")
 }
@@ -11,13 +11,8 @@ cov <- covr::package_coverage(
 )
 print(cov)
 
-out_file <- "r-coverage.xml"
-covr::to_cobertura(cov, filename = out_file)
-
-root <- normalizePath(getwd(), winslash = "/", mustWork = TRUE)
-xml <- paste(readLines(out_file, warn = FALSE), collapse = "\n")
-xml <- gsub("\\\\", "/", xml)
-xml <- gsub(root, ".", xml, fixed = TRUE)
-writeLines(xml, out_file)
+out_file <- "r-coverage.json"
+# to_codecov() is the JSON body that covr::codecov() uploads.
+writeLines(covr:::to_codecov(cov), out_file)
 
 message("Wrote ", out_file)
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -6,32 +6,28 @@ on:
     branches: [main, master]
 
 jobs:
-  R-CMD-check:
+  coverage-tests:
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        test-suite: [ R_coverage, JS_coverage, CRAN ]
+        test-suite: [ R_coverage, JS_coverage ]
 
     name: ${{ matrix.test-suite }}
-    env: 
+    env:
       TEST_SUITE: ${{ matrix.test-suite }}
       GITHUB_PAT: ${{ secrets.PAT_GITHUB }}
       GH_ACTION: "ENABLED"
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
     steps:
-      - uses: actions/checkout@v3
-      - name: install and update texlive
-        run: /usr/bin/sudo DEBIAN_FRONTEND=noninteractive apt update -y -qq
-      - run: /usr/bin/sudo DEBIAN_FRONTEND=noninteractive apt install tidy texlive texlive-fonts-extra -y
+      - uses: actions/checkout@v4
       - uses: r-lib/actions/setup-r@v2
       - uses: r-lib/actions/setup-r-dependencies@v2
 
       - name: install package
         run: R CMD INSTALL .
 
-      - name: git config user.name 
+      - name: git config user.name
         run: git config --global user.name "GitHub Actions"
 
       - name: git config user.email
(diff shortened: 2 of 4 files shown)


tdhock:
@ANAMASGARD please review?

.github/scripts/save-r-coverage.R
```diff
@@ -0,0 +1,23 @@
...
+
+cov <- covr::package_coverage(
+  quiet = FALSE,
+  type = "tests",
+  test_files = "tests/testthat.R"
+)
+print(cov)
+
+out_file <- "r-coverage.xml"
+covr::to_cobertura(cov, filename = out_file)
```
ANAMASGARD: @Nishita-shah1  could you preserve the previous R coverage measurement here instead of switching to Cobertura? Codecov now sees 9494 lines instead of 8933 (+561), even though this PR does not change production code, which is why codecov/project still reports -0.10%.

Please generate the Rlang JSON equivalent to the previous covr::codecov() report and save that as the artifact, while keeping the new combined upload.

.github/workflows/tests.yaml
```diff
@@ -42,44 +38,112 @@ jobs:
...
+          test -f coverage-istanbul.json
+
+      - name: Upload R and JS coverage together
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: r-coverage.xml,coverage-istanbul.json
+          disable_search: true
+          fail_ci_if_error: true
+          verbose: true
```
ANAMASGARD, line 122:
```suggestion
      - name: Verify both coverage reports exist
        run: |
          test -s r-coverage.json
          test -s coverage-istanbul.json

      - name: Upload R and JS coverage together
        uses: codecov/codecov-action@v5
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          files: r-coverage.json,coverage-istanbul.json
          disable_search: true
          fail_ci_if_error: true
          verbose: true
```

ANAMASGARD: After changing the R report back to the previous Rlang JSON measurement, could you update the verification and combined upload here too? Please also use test -s so empty coverage reports fail.

.github/workflows/tests.yaml
```diff
@@ -42,44 +38,112 @@ jobs:
...
+      - name: Run R tests and save coverage
+        if: matrix.test-suite == 'R_coverage'
+        run: Rscript .github/scripts/save-r-coverage.R
+
+      - name: Upload R coverage artifact
+        if: matrix.test-suite == 'R_coverage'
+        uses: actions/upload-artifact@v4
+        with:
+          name: r-coverage
+          path: r-coverage.xml
```
ANAMASGARD: Please also update the earlier R artifact path from r-coverage.xml to r-coverage.json

tests/testthat/test-renderer3-knit-print-pdf.R
```diff
@@ -1,4 +1,5 @@
 acontext("render animint in pdf")
+skip_if_not(nzchar(Sys.which("pdflatex")), "pdflatex not available")
```
ANAMASGARD, line 2: @Nishita-shah1  after removing Texlive from the coverage jobs, this test skips there. The CRAN job has Texlive, but build.sh uses filter="compiler", so this renderer test is not selected there either.

could we please keep this test running in at least one CI job, so the coverage test scope remains consistent with master?

ANAMASGARD (commented):
@Nishita-shah1  the combined coverage upload approach looks correct for #254. I left a few inline comments 
after these changes, could you please verify one failure case where either coverage job fails and the Codecov upload is skipped, followed by a docs-only/no-code-change run showing 0.00% project coverage change? that would confirm #254 end-to-end.

.github/workflows/tests.yaml
```diff
@@ -42,44 +38,112 @@ jobs:
...
+        run: R CMD INSTALL .
+
+      - name: git config user.name
+        run: git config --global user.name "GitHub Actions"
+
+      - name: git config user.email
+        run: git config --global user.email toby.hocking@r-project.org
+
+      - name: run CRAN build
+        run: bash build.sh
```
ANAMASGARD, line 152: could you run the PDF renderer test here, since this is the job where Texlive is installed? currently the coverage jobs skip it, and build.sh only runs filter="compiler".
```suggestion
      - name: run PDF renderer test
        run: Rscript -e 'testthat::test_file("tests/testthat/test-renderer3-knit-print-pdf.R")'

      - name: run CRAN build
        run: bash build.sh
```

.github/workflows/tests.yaml
```diff
@@ -42,44 +38,115 @@ jobs:
...
+        run: git config --global user.name "GitHub Actions"
+
+      - name: git config user.email
+        run: git config --global user.email toby.hocking@r-project.org
+
+      - name: run PDF renderer test
+        run: Rscript -e 'testthat::test_file("tests/testthat/test-renderer3-knit-print-pdf.R")'
+
+      - name: run CRAN build
+        run: bash build.sh
```
ANAMASGARD, line 152: sorry, my earlier suggestion missed the normal test setup. test_file() sources the helpers directly, but animint2 is not loaded, so toRGB() is unavailable. could we load the same packages as tests/testthat.R before running this test?
```suggestion
      - name: run PDF renderer test
        run: Rscript -e 'library(testthat); library(animint2); library(XML); test_file("tests/testthat/test-renderer3-knit-print-pdf.R")'

      - name: run CRAN build
        run: bash build.sh
```

ANAMASGARD:
Sir @tdhock  the PR now looks good can you please give it a last review before merge .

tdhock:
please resolve conflicts

Nishita-shah1:
Hi @tdhock , i have resolved conflicts

tdhock:
please fix?
```
* checking package subdirectories ... NOTE
Problems with news in ‘NEWS.md’:
  Cannot extract version info from the following section titles:
    Changes in development (PR#339)
```

Nishita-shah1:
Hi @tdhock, everything is fixed now. Could you please review it

tests/testthat/test-renderer3-knit-print-pdf.R
tdhock: please undo this addition, since it is unrelated to PR subject.

tdhock (changes requested):

NEWS.md
```diff
@@ -6,10 +6,9 @@
 
 - Multi-line facet strip labels: `facet_grid()` and `facet_wrap()` now render each facet variable on its own line, matching ggplot2 (issue #262).
 
-# Changes in version 2026.6.5 (PR#336)
+# Changes in version 2026.8.18 (PR#344)
 
-- `geom(showSelected.legend=FALSE)` is the canonical opt-out for legend-driven auto-injection of showSelected, including while keeping explicit showSelected variables. (Fixed #333)
-- Breaking API cleanup: PR #292 `geom(showSelected=character())` no longer opts out; use `showSelected.legend=FALSE` instead.
+- Codecov now uploads R and JavaScript coverage together, and only after both coverage jobs succeed, so a failed JS (or R) job cannot become an incomplete project-coverage baseline.
```
tdhock: can you please revise this to explain what CI jobs are defined now in tests.yaml? 
is it true that it defines R_coverage and JS_coverage, and a new upload-coverage which depends on the other two?

.github/workflows/tests.yaml
```diff
@@ -42,44 +38,115 @@ jobs:
...
+        run: R CMD INSTALL .
+
+      - name: git config user.name
+        run: git config --global user.name "GitHub Actions"
+
+      - name: git config user.email
+        run: git config --global user.email toby.hocking@r-project.org
+
+      - name: run PDF renderer test
+        run: Rscript -e 'library(testthat); library(animint2); library(XML); test_file("tests/testthat/test-renderer3-knit-print-pdf.R")'
```
tdhock: please delete, should be un-necessary

tdhock (changes requested):

Nishita-shah1:
hi @tdhock , done all the changes

tdhock (approved):
looks good thanks

PR #346, opened 2026-09-06
Added regression test for geom_text interactive updates (#345)

### LINKED ISSUE #345: setMultilineText() breaks geom_text interactive updates
https://github.com/tdhock/2026-08-apprentissage/blob/main/figure-sur-sous-apprentissage-test-neighbors.R breaks with new code from #247 which does not work with interactive updates to geom_text() elements. @ANAMASGARD can you please investigate and create a simpler test case that fails?
After clicking 40 voisins (neighbors) it should look like this:

Image

but using setMultilineText(), the geom_text at y=sélection* is not updated to the correct x position.
Reverting to .text() fixes this, so I will push that fix to master.

ANAMASGARD:
## Summary
- Added a minimal Chromote renderer test for #345: `geom_text` with a stable D3 `key`, `showSelected`, and non-zero `duration` must move its **rendered** position after clicking a `geom_tallrect`.
- Protects Toby’s existing `.text()` fix on master from regressing if `setMultilineText()` is re-enabled for `geom_text`.

Closes #345

### Local result
- Local Chromote run of `test-renderer3-geom-text-interactive-updates.R` 
image

### DIFF
--- a/tests/testthat/test-renderer3-geom-text-interactive-updates.R
+++ b/tests/testthat/test-renderer3-geom-text-interactive-updates.R
@@ -1,0 +1,33 @@
+acontext("geom_text interactive updates - issue #345")
+library(animint2)
+text.df <- data.frame(
+  item = c("a", "b"),
+  x = c(1, 9),
+  y = 1,
+  label = "marker",
+  key = "label")
+rect.df <- data.frame(
+  item = c("a", "b"),
+  xmin = c(0.6, 8.6),
+  xmax = c(1.4, 9.4),
+  id = c("select_a", "select_b"))
+viz <- animint(
+  plot1 = ggplot() +
+    theme_animint(width = 400) +
+    scale_x_continuous(limits = c(0, 10)) +
+    geom_text(aes(x, y, label = label, key = key),
+              showSelected = "item", data = text.df) +
+    geom_tallrect(aes(xmin = xmin, xmax = xmax, id = id),
+                  clickSelects = "item", data = rect.df),
+  duration = list(item = 500),
+  first = list(item = "a"))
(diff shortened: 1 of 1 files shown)


tests/testthat/test-renderer3-geom-text-interactive-updates.R
```diff
@@ -0,0 +1,33 @@
...
+  first = list(item = "a"))
+info <- animint2HTML(viz)
+get_geom_text_center_x <- function() {
+  get_element_bbox("svg#plot_plot1 g.geom1_text_plot1 text.geom")$center_x
+}
+x_start <- get_geom_text_center_x()
+clickID("select_b")
+Sys.sleep(1)
+test_that("geom_text x position updates after showSelected changes", {
+  expect_gt(get_geom_text_center_x() - x_start, 50)
```
tdhock: can you please avoid magic numbers like 50 here?

can you instead test for equality with another geom that does not have showSelected?

tdhock (changes requested):

tests/testthat/test-renderer3-geom-text-interactive-updates.R
ANAMASGARD: @tdhock  Thanks for review . On multiline: my bad the old test only checked compiler output, not browser rendering
I added a Chromote test that fails on master with .text(). On the interactive test: I replaced the 50px threshold with equality against a static reference geom_text at x=9, plus a before-click check that the interactive label starts to the left.

tdhock: ok that makes sense

